PR #7, opened 2023-01-09
PDFT quadrature small grid case blksize align 

MatthewRHermes:
cf. PySCF commit fb038b40eef6fb76f84f1801403d4a969e661d12

new assertion in pyscf/dft/numint.py line 2652.


MatthewRHermes:
This is in PR #8

PR #10, opened 2023-02-02
include user-defined cint.h directory for cmake

hebrewsnabla:
Looks like https://github.com/MatthewRHermes/mrh/pull/15 is not applied after migrating to pyscf-forge. Applied in this PR.


MatthewRHermes:
Commentary: I think PR #3 made this necessary by removing my copy-pasted code and grabbing things instead from PySCF; it's the PySCF header files that need to see cint.h; not nr_numint.c itself.

PR #13, opened 2023-02-27
Linearized Pair-Density Functional Theory

matthew-hennefarth:
Implementation of Linearized Pair-Density Functional Theory (as described here https://doi.org/10.26434/chemrxiv-2023-7d0gv). 

This is a new multi-state PDFT method which takes a different approach/philosophy for computing off-diagonal coupling elements. As such, the code is placed in its own lpdft.py file. In general, it uses the mcpdft.multi_state() function but uses the `method="lin"` argument (whereas CMS-PDFT uses `method="cms"`).

Hybrid L-PDFT is performed using `hybrid_kernel` until the hybrid `veff_1`, `veff_2` is sorted out. 

Tests for lih with 2 and 4 states is provided as well. 

A simple example is given and an update to the mcpdft doc is also added.


matthew-hennefarth:
Updated L-PDFT so that hybrid functionals (like tPBE0) can be used directly with the kernel function and deleted the hybrid_kernel function. Added a kwarg to make it clear in get_pdft_veff that the derivatives do not have the mcwfn component and is purely the functional term (this was a discussion with Matt Hermes)

Issue #19, opened 2023-05-02
setup.py broke again


MatthewRHermes:
See the thread in PR #18. I do not think the CI failures there have anything to do with the PR itself.

[Experimentation](https://github.com/MatthewRHermes/pyscf-forge/tree/debug_ci) suggests that the dependencies indicated by setup.py are, for some reason, now placed in a location that the sys.path inside of setup.py cannot see:
```
Requirement already satisfied: pyscf in /opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages (from pyscf-forge==1.0.0) (2.2.1)
Requirement already satisfied: numpy in /opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages (from pyscf-forge==1.0.0) (1.21.6)
Requirement already satisfied: scipy!=1.5.0,!=1.5.1 in /opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages (from pyscf->pyscf-forge==1.0.0) (1.7.3)
Requirement already satisfied: h5py>=2.7 in /opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages (from pyscf->pyscf-forge==1.0.0) (3.8.0)
```
whereas `print (sys.path)` inside of `setup.py:CMakeBuildExt.build_cmake` produces:
```
['/home/runner/work/pyscf-forge/pyscf-forge', '/opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages/pip/_vendor/pyproject_hooks/_in_process', '/tmp/pip-build-env-ofuuo5v2/site', '/opt/hostedtoolcache/Python/3.7.16/x64/lib/python37.zip', '/opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7', '/opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/lib-dynload', '/tmp/pip-build-env-ofuuo5v2/overlay/lib/python3.7/site-packages', '/tmp/pip-build-env-ofuuo5v2/normal/lib/python3.7/site-packages']
```
I can maybe think of some horrible hodge-podge way to "fix" this in run_ci.sh, but really I would appreciate some help figuring out the "right" way to fix it.

MatthewRHermes:
The utterly bizarre thing to me is that it's so close:
```
[...]

PR #18, opened 2023-05-01
PySCF PR #1623 compatibility

MatthewRHermes:
PySCF PR #1623 made symmetry FCI solvers unable to interpret a 3d ndarray CI vector. Fix this and remove a certain unit test case which was always undefined behavior.

This presumes PySCF PR #1721

PR #21, opened 2023-05-02
pyscf PR #1623 compatibility and np.int deprecation

MatthewRHermes:
PySCF PR #[1623](https://github.com/pyscf/pyscf/pull/1623) made symmetry FCI solvers unable to interpret a 3d ndarray (`shape=(n,*,*)`) as a list of _n_ CI vectors, which was used by one unit test. Here, those are explicitly cast to a list of 2d ndarrays. Also remove a certain unit test case which was always undefined behavior. This presumes PySCF PR #[1721](https://github.com/pyscf/pyscf/pull/1721). See issue #19 and PR #20 concerning the inevitable CI failure. I redid this from PR #18 after getting sick of my master branch git log constantly having a bunch of meaningless "merge master from origin" lines in it.

### DIFF
--- a/pyscf/grad/test/test_grad_mspdft.py
+++ b/pyscf/grad/test/test_grad_mspdft.py
@@ -81,11 +81,11 @@ class KnownValues(unittest.TestCase):
 
     def test_offdiag_response_sanity (self):
         for mcs, stype in zip (get_mc_list (), ('nosymm','symm')):
-         if 'no' not in stype:
-            continue
-            # TODO: redesign this test case. explicitly mixing CI vectors
-            # like this is undefined behavior
          for mca, atype in zip (mcs, ('nomix','mix')):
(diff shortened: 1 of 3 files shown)


MatthewRHermes:
To explain the changes to test_grad_mspdft: prior to PySCF PR #1623, the `contract_2e` methods of the symmetry-adapted FCI solvers didn't actually use point group symmetry, and it was therefore possible to pass a symmetry-broken CI vector to these methods and get the no-symmetry result. This is obviously silly, and you shouldn't do this. The reason this even came up was that I was trying to test certain intermediate functions for MS-PDFT gradients in the case of both `state_average` and `state_average_mix` type calculations. However, I didn't realize at the time that we hadn't even really defined what was a `state_average_mix` version of CMS-PDFT even meant.

PR #22, opened 2023-05-03
CI/CD run unit tests

MatthewRHermes:
The CI should probably run the unit tests in addition to compiling. Some tests may also be too long to be run in CI, and some were not originally in the right format, etc., to be used with pytest. Let's fix that.

~~(The failed CI here is related to issue #19 and PR #20).~~


MatthewRHermes:
Any comment @sunqm or @fishjojo? Are these tests too burdensome? Is there something else I should have the github actions do?

fishjojo (approved):

Issue #29, opened 2023-05-29
SAMix PDFT Error (Single State)


rebeccadanto:
Getting the following error when running mcpdft on a single root: 
<pre><code>Traceback (most recent call last):
  File "/home/rdanto/.local/lib/python3.8/site-packages/pyscf/mcscf/addons.py", line 1188, in __getitem__
    return self._data[key]
TypeError: 'int' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "new_script.py", line 101, in <module>
    mc.kernel()
  File "/home/rdanto/apps/pyscf_ext/pyscf-forge/pyscf/mcpdft/mcpdft.py", line 488, in kernel
    self.compute_pdft_energy_(otxc=otxc, grids_attr=grids_attr,
  File "/home/rdanto/apps/pyscf_ext/pyscf-forge/pyscf/mcpdft/mcpdft.py", line 482, in compute_pdft_energy_
    self.e_tot, self.e_ot = self.energy_tot (mo_coeff=self.mo_coeff, ci=self.ci)
  File "/home/rdanto/apps/pyscf_ext/pyscf-forge/pyscf/mcpdft/mcpdft.py", line 680, in energy_tot
    e_tot, e_ot = energy_tot (self, mo_coeff=mo_coeff, ot=ot, ci=ci,
  File "/home/rdanto/apps/pyscf_ext/pyscf-forge/pyscf/mcpdft/mcpdft.py", line 72, in energy_tot
    casdm1s = mc.make_one_casdm1s (ci, state=state)
  File "/home/rdanto/apps/pyscf_ext/pyscf-forge/pyscf/mcpdft/_dms.py", line 76, in make_one_casdm1s
    return fcisolver.make_rdm1s (ci, ncas, nelecas)
  File "/home/rdanto/.local/lib/python3.8/site-packages/pyscf/mcscf/addons.py", line 1360, in make_rdm1s
    dm1a, dm1b = self.states_make_rdm1s (ci0, norb, nelec, link_index=link_index, **kwargs)
  File "/home/rdanto/.local/lib/python3.8/site-packages/pyscf/mcscf/addons.py", line 1354, in states_make_rdm1s
    for dm1s in self._collect ('make_rdm1s', ci0, norb, nelec, link_index=link_index, **kwargs):
  File "/home/rdanto/.local/lib/python3.8/site-packages/pyscf/mcscf/addons.py", line 1286, in _collect
[...]

PR #30, opened 2023-05-30
Samix pdft nroots=1 edge case

MatthewRHermes:
Address issue #29. _dms.py now inspects the class identity of mc.fcisolver regardless of the value of mc.fcisolver.nroots, which is the way it should have been done in the first place. N.B. the reason I update the attributes of the base class in the StateAverageFCISolver case but not in the StateAverageMixFCISolver case is that only in the former case am I instantiating a new method instance; in the latter case it should already have been fully initialized.

ETA: compute_pdft_energy_ now also always checks for StateAverageMCSCFSolver, even if nroots==1.


dking072:
With the current branch, mc.e_states returns [None] with sa_mix and 1 root with pdft

dking072:
Current commit fixes the issue 👍

fishjojo:
please fix conflicts

PR #32, opened 2023-06-01
_dms.py args -> kwargs ; _mcscf_env clean

MatthewRHermes:
Fix a bug where make_one_casdm* takes ci as an arg despite it sometime being called with ci as a kwarg. (kwargs are safer here). Also automate more variable name reassignments in optimize_mcscf_ so that its easier to apply PDFT to MC-SCF subclasses that happen to have different kernel return signatures.


MatthewRHermes:
Closing this for now because I guess I didn't quite understand the issue.

Issue #40, opened 2023-10-18
PySCF PR #1859 incompatible with MC-PDFT


MatthewRHermes:
PySCF's [PR #1859](https://github.com/pyscf/pyscf/pull/1859) changed `__init__` methods of several classes which is causing inconsistency with MC-PDFT here. Hopefully an easy fix.

sunqm:
Thanks for the fixes of your PR!

PR #41, opened 2023-10-18
Compatibility w/ PySCF #1859 (issue #40)

MatthewRHermes:
Update some class initializations to correspond to the PySCF MCSCF refactor in PySCF PR #1859. This is the minimal necessary change to pass existing unit tests.


matthew-hennefarth:
Thanks for this!

MatthewRHermes:
SCF convergence tolerances may need to be increased for test_lpdft.py, or else validation thresholds should be loosened.

PR #43, opened 2024-01-26
L-PDFT Nuclear Gradients

matthew-hennefarth:
Analytic nuclear gradients for linearized pair-density functional theory based on a SA-CASSCF wave function. Theory and details as described in [**arXiv**, 2401.12933](https://arxiv.org/abs/2401.12933). 

Note, currently does not support density fit integrals, mixed solvers comprised of states from different spin states, or hybrid functionals. The mcpdft/README.md has been updated and tests have also been included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,7 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# IDEs
+.idea
+.vscode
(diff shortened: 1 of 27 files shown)


sunqm:
@matthew-hennefarth Will you submit the work to peer reviewed journals? If so, should we wait for the DOI and cite that in the code?

matthew-hennefarth:
> Will you submit the work to peer reviewed journals? If so, should we wait for the DOI and cite that in the code?

Yes, it is submitted to a peer reviewed journal. We can wait for the DOI and add the citation once it is published.

matthew-hennefarth:
@sunqm added the DOI for L-PDFT Analytical gradients and fixed all merge conflicts.

PR #47, opened 2024-05-08
M3SOSCF Implementation

LinusBDittmer:
This is an implementation of the Markovian Multiagent Monte-Carlo SOSCF based on CIAH. It is a powerful, albeit slow SCF algorithm that enables near-guaranteed convergence even in extremely difficult systems.

Publication: [https://doi.org/10.1063/5.0159737](https://doi.org/10.1063/5.0159737)

See also [PR#2208 in pyscf/pyscf](https://github.com/pyscf/pyscf/pull/2208).


pyscf/scf/hf.py
matthew-hennefarth, line 1: This file is the same as that in the main [PySCF repository](https://github.com/pyscf/pyscf/blob/master/pyscf/scf/hf.py). You should not copy those files into this repository. Instead, you should import this file using the appropriate method from the main PySCF directory.

pyscf/soscf/test/test_m3soscf.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
+# Copyright 2014-2019 The PySCF Developers. All Rights Reserved.
```
matthew-hennefarth, line 2: Update copyright to 2024

pyscf/soscf/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
```
matthew-hennefarth, line 1: Update Copyright to 2024

pyscf/soscf/m3soscf.py
```diff
@@ -0,0 +1,858 @@
+'''
```
matthew-hennefarth, line 1: Add copyright comment banner.

pyscf/soscf/sigma_utils.py
```diff
@@ -0,0 +1,32 @@
+'''
```
matthew-hennefarth, line 1: Add copyright banner

matthew-hennefarth (changes requested):

matthew-hennefarth:
Can you please add an example of this functionality.

pyscf/scf/hf.py
LinusBDittmer: I added a decorated to the SCF base class in hf.py that allows for the creation of an M3SOSCF object analogously to the CIAH implementation. Should I add this differently, and if so, what specifications should I follow?

LinusBDittmer:
An example has been added in examples/scf. Copyright banners have been updated/added.

pyscf/soscf/__init__.py
matthew-hennefarth, line 1: This file should be deleted since it will cause import clashes later on down the line with that in the main branch.

pyscf/scf/hf.py
matthew-hennefarth: I believe so otherwise there will be import clashes. That is, when someone does 

```python
import pyscf.scf.hf
```
or 
```python
from pyscf.scf import hf
```

It is somewhat undefined which `hf.py` to import, that from pyscf or from pyscf-forge. I am not sure the best way of doing this to get the desired result. This is because pyscf-forge is built such that it doesn't really modify the internal structure of classes from pyscf. One could get around this by constructing a new SCF object that inherits the SCF object from pyscf/pyscf/scf/hf.py but then adds on the `m3soscf` method. This would again have to be implemented such that there are no import conflicts from the main repo though.

matthew-hennefarth: One possibility, but this is more of a general design decision for pyscf-forge, is to place all of your new code in a subfolder m3soscf. This would prevent import clashes as the hf.py would be in pyscf.m3soscf.scf.hf instead. But in general, I am not sure of the correct way of doing this for this project.

LinusBDittmer: I suggest that I remove hf.py for now, as there still is a straightforward constructor without it instead. If an M3SOSCF implementation is merged into pyscf proper, the decorator can still be added.

LinusBDittmer:
soscf/__init__.py as well as scf/hf.py have been removed. Since there are no other files in scf/, I have removed the directory entirely.

matthew-hennefarth:
It should be good so long as all tests pass, @sunqm can you please approve for the workflows?

LinusBDittmer:
I still had the old import statements from developing inside a forked repo of pyscf proper, I changed these to work for pyscf-forge.

sunqm:
> It should be good so long as all tests pass, @sunqm can you please approve for the workflows?

Thank you for reviewing the PR code!

LinusBDittmer:
@matthew-hennefarth Thank you for the review. @sunqm Is there anything I can do, any information I can provide, regarding possible discussion on whether this feature will be merged into the main branch of PySCF?

PR #49, opened 2024-05-30
Contributing guideline

sunqm:


fishjojo:
we could link this in the doc, and also list the current available modules so that people's work can be recognized.

sunqm:
@gkc1000 could you edit the guideline with the points you mentioned in the meeting? You have the permission to edit the file directly.

gkc1000:
Yes, I will later today

PR #53, opened 2024-06-17
Numpy 2.0 compatibility

MatthewRHermes:
remove np.string_


MatthewRHermes:
I think the failed test is a fluke (tolerance is 5e-6; delta is 6e-6; method is CASCI-PDFT which is highly susceptible to rounding errors)

sunqm (approved):

sunqm:
I agree. How about loosening the threshold a little bit?

MatthewRHermes:
sure

PR #54, opened 2024-06-18
Add a function in tools_td.py file in sftda module to calculate the <S^2> for excited states.

Haskiy:
sftda.tools_td.spin_square fucntion serves for <S^2> calculations of excited states. The keywards extype and tdtype control excitation classes and tda/ tddft, respectively. extype=0,1,2 serves for spin flip up, spin flip down and spin conserving excitations. tdtype='TDA', 'TDDFT' serves for tda or tddft results. And this function can be used at logger system to give the information of <S^2> for excited states like gaussian, qchem in the future. There is one point that this function can calculate <S^2> of one excited state according to the input transition vector, which can be modified if the function is used in the logger system.
Besides, a bug for spin flip TDA, collinear_samples can not be changed, is fixed also.


examples/sftda/01-spin_square.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env/python
+
+'''
+Test the spin_square calculation tool in sftda.tools_td.py.
+'''
+
+from pyscf import gto,dft,sftda,tdscf
+from pyscf.sftda import tools_td
+
+# ToDo : add the spin flip TDDFT parts.
```
MatthewRHermes, line 10: Can you elaborate on this? Is this function missing terms?

Haskiy: Not this function missing terms. There isn't a good davidson or other solvers to solve the spin flip tddft matrix, and the sftda module  can be used in spin flip tda using the davidson solver in pyscf only. For the spin flip tddft, we recommend to solve the whole spin flip tddft matrix by the help of get_ab_sf fucntion, which has been applied in the sftda module,  until we or other people find a good solver. So the sftda module can not give a spin flip tddft calculation results directly, that need users to construct a calculation result to give tools_td.spin_square() and obtain the <S^2>.

examples/sftda/01-spin_square.py
```diff
@@ -0,0 +1,68 @@
...
+
+#
+# 1. <S^2> for spin flip up TDA
+#
+mftd1 = sftda.TDA_SF(mf)
+mftd1.extype=0
+mftd1.nstates = 5
+mftd1.kernel()
+
+# extype serves for controling spin excitation type in the spin_square function.
```
sunqm, line 34: Please add more comments to this attribute, such as how the 0, 1, 2 corresponds to the spin flip cases.

tberkel:
Really the `extype` attribute should probably be improved at the class
level. Using simple integers as flags is confusing and not used much
anywhere else in PySCF. A more descriptive way to specify the spin flip
cases would be better.

examples/sftda/01-spin_square.py
Haskiy: OK, I will add these comments to this attribute.

Haskiy:
The `extype` 0,1,2 are used to indicate spin flip up, spin flip down and spin conserving excitation, respectively. The three integers does make users confused. The number 0 and 1 for attribute 'extype' have been used at sftda module to control spin flip excitation classes, so I use number 2 to indicate spin conserving case in the spin_square fucntion. Recently, I used BDF (Beijing Density Functional) package, in which 1,0,-1 are used to control the excitation class.  It's worth considering your advice that use spin_square fucntion as a class function. @tberkel

sunqm (approved):

PR #59, opened 2024-08-05
Local ISDF module (with k-sampling) for PBC mean-field calculation.

NingZhang1:
1. Local Interpolative separable density fitting module for PBC mean-field calculation.
2. Supporting robust fitting, k-point, MPI.
3. L-ISDF supports ao2mo but ao2mo_7d has not been implemented yet.


fishjojo:
I think we can just keep the local version.

sunqm:
Can `scipy.linalg.cho_solve` be used to replace these low-level Cholesky Lapack functions?

NingZhang1:
> Can `scipy.linalg.cho_solve` be used to replace these low-level Cholesky Lapack functions?

Of course, if needed.

PR #62, opened 2024-08-13
Local ISDF module (with k-sampling) for PBC mean-field calculation.

NingZhang1:
1.Local Interpolative separable density fitting module for PBC mean-field calculation.
2.Supporting robust fitting, k-point, MPI.
3.L-ISDF supports ao2mo but ao2mo_7d has not been implemented yet.


sunqm:
The code in python seems only using the np.fft and the fft/ifft code from `pyscf.lib`. Is the fftw wrapper `fft.c` still needed?

NingZhang1:
> The code in python seems only using the np.fft and the fft/ifft code from `pyscf.lib`. Is the fftw wrapper `fft.c` still needed?

The calculation of V matrix (which is the counterpart of auxiliary Coulomb matrix in DF) heavily depends on fft/ifft (please see pbc_isdf_V.c) and is computationally demanding.  It fftw3.h is not allowed to be included in pyscf-forge, I can rewrite all the involved code using np.fft/np.ifft or other python wrapper for fftw3 to see whether this will influence the performance.

gkc1000:
I think the normal way is to at least have numpy/scipy (the standard pyscf
dependencies) as a fallback.

NingZhang1:
Got it! Thanks for suggesting that.

gkc1000 ***@***.***> 于2024年8月16日周五 08:47写道：

> I think the normal way is to at least have numpy/scipy (the standard pyscf
> dependencies) as a fallback.
>

sunqm:
Please also clean up the style issues. You can run the flake8 locally `flake8 --config /path/to/pyscf-forge/.flake8 /path/to/pyscf-forge`

examples/isdf/00-RHF_diamond.py
```diff
@@ -0,0 +1,121 @@
...
+                    mf.max_cycle = 32
+                    mf.conv_tol = 1e-7
+                    pbc_isdf_info.direct_scf = mf.direct_scf
+                    mf.kernel()
+                    t2 = (lib.logger.process_clock(), lib.logger.perf_counter())
+                    print(isdf_jk._benchmark_time(t1, t2, 'scf_isdf', pbc_isdf_info))
+                        
+                    del mf
+                    del pbc_isdf_info
+                exit(1)
```
matthew-hennefarth, line 121: Is there a reason why you `exit(1)` in all of the examples? This signifies an error has occurred which should not be the case if it reaches the end of the code.

pyscf/isdf/__init__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
```
matthew-hennefarth, line 1: Please update the copyright to `2014-2024` here and for the rest of the files.

matthew-hennefarth:
Will you be providing any tests for this module?

examples/isdf/00-RHF_diamond.py
NingZhang1: Thanks for pointing out this issue. It is added just because I do not want run all the for-loops when I am testing the code. It should not appear here and I will comment this out.

pyscf/isdf/__init__.py
NingZhang1: Thanks, I will change all of them.

NingZhang1:
> Will you be providing any tests for this module?

I have some test files within my local branch of ISDF module, I can add them back if needed.

PR #64, opened 2024-09-07
Energy Decomposition Analysis adopted for general MCSCF 

JangidBhavnesh:
Previously EDA is hardcoded for the CASSCF wavefunction only. I added the functionality to do this analysis for the DMRG-PDFT as well.

### DIFF
--- a/pyscf/mcpdft/mcpdft.py
+++ b/pyscf/mcpdft/mcpdft.py
@@ -289,7 +289,7 @@ def get_energy_decomposition(mc, mo_coeff=None, ci=None, ot=None, otxc=None,
         e_1e = []
         e_coul = []
         e_otxc = []
-        e_ncwfn = [] 
+        e_ncwfn = []
         for ix in range(nroots):
             row = _get_e_decomp(mc, mo_coeff, ci, ot, state=ix)
             e_1e.append(row[0])
--- a/pyscf/mcpdft/mcpdft.py
+++ b/pyscf/mcpdft/mcpdft.py
@@ -289,8 +289,7 @@ def get_energy_decomposition(mc, mo_coeff=None, ci=None, ot=None, otxc=None,
         e_1e = []
         e_coul = []
         e_otxc = []
-        e_ncwfn = []
-        
+        e_ncwfn = [] 
         for ix in range(nroots):
             row = _get_e_decomp(mc, mo_coeff, ci, ot, state=ix)
             e_1e.append(row[0])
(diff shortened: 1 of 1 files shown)


matthew-hennefarth:
LGTM. Do you know if any of the tests actually test this feature? Perhaps we could add one if not.

JangidBhavnesh:
We don't have any test for this functionality. 

I will add a unit test. Here, the sum of EDA components would be equal to the MCPDFT energy for the given state.

matthew-hennefarth:
> Here, the sum of EDA components would be equal to the MCPDFT energy for the given state.

Not necessarily, I believe. Doesn't it also include more than just the Coulomb interaction, such as CAS exchange? As in `e_ncwfn` includes full 2e active space contribution if I read the code correctly. Also, for hybrid funcs it might disagree.

JangidBhavnesh:
Sorry, there is a test for energy_decomposition. Though, whatever I have changed is passing the test.
https://github.com/pyscf/pyscf-forge/blob/master/pyscf/mcpdft/test/test_mcpdft.py

Also, energy_decomposition is not implemented for the hybrid functionals. 
https://github.com/pyscf/pyscf-forge/blob/master/pyscf/mcpdft/mcpdft.py#L274

However, it won't be that difficult to implement it. It can be represented as
$E_{OT} = \lambda (\frac{1}{2}g_{pqrs}D_{pqrs}) + (1-\lambda ) E[\rho, \Pi]$
Where it can further be decomposed into exchange and correlation.

JangidBhavnesh:
> Not necessarily, I believe. Doesn't it also include more than just the Coulomb interaction, such as CAS exchange? As in `e_ncwfn` includes full 2e active space contribution if I read the code correctly. Also, for hybrid funcs it might disagree.

Yes, you are right. MCPDFT energy doesn't include the e_ncwfn. 
$E_{MCPDFT} = sum(E_{Decomposition}[:3  or  :4])$ (depends on split_xc)

matthew-hennefarth:
> However, it won't be that difficult to implement it. It can be represented as E O T = λ ( 1 2 g p q r s D p q r s ) + ( 1 − λ ) E [ ρ , Π ] Where it can further be decomposed into exchange and correlation.

You don't have to even do that. The MC-PDFT energy should actually be $E^{PDFT} = E^{1e} + Coul + \lambda E^{ncwfn} + (1-\lambda)E^{ot}$. If $\lambda = 0$, it will still work.

MatthewRHermes (approved):
Thanks for cleaning up my old spaghetti!

PR #66, opened 2024-09-09
check integer correctly to account for pyscf #2377

MatthewRHermes:
`isinstance (*, int) -> isinstance (*, (int, np.integer))`

### DIFF
--- a/pyscf/mcpdft/_libxc.py
+++ b/pyscf/mcpdft/_libxc.py
@@ -16,6 +16,7 @@
 import numpy as np
 from pyscf.dft.libxc import XC_ALIAS, XC_CODES, XC_KEYS
 from pyscf.dft.libxc import hybrid_coeff, rsh_coeff
+from pyscf import lib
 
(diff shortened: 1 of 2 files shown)


fishjojo:
There is a function `isinteger` in `lib.misc` that we can use.

MatthewRHermes:
Changed the title because on closer inspection I don't think this is a numpy 2.0 thing

MatthewRHermes:
OK, figured out why this became a problem for me now: [pyscf #2377](https://github.com/pyscf/pyscf/pull/2377) changed the data type of `libxc.XC_CODES.values ()` from native Python integer to `numpy.intc` in order to interface with libxc functional keyword parsing. The reason the github actions isn't catching this is, as in #67, we need the PySCF 2.7 release in order to see the effects of [pyscf #2377](https://github.com/pyscf/pyscf/pull/2377).

sunqm (approved):

PR #60, opened 2024-08-09
Trexio interface

sunqm:
Regarding the feature request https://github.com/pyscf/pyscf/issues/2308

PR #68, opened 2024-09-19
Implemented the ECP part and added several missing factors in the TREXIO interface.

kousuke-nakano:
@matthew-hennefarth 

Sorry for my late response. Today, I have implemented lines to write ECP information to a TREXIO file. Please have a look at my pull request.

The ECP part works. However, a problem I found is that the `cc-pVXZ` basis sets accompanied with `ccECP`, which are often used for QMC calculations, did not pass the following assertion line in `trexio.py`.

>       assert all(mol._bas[:,gto.NCTR_OF] == 1)

Would you give me a clue for solving this problem?

#60

### DIFF
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -17,7 +17,6 @@ import pyscf
 from pyscf import lib
 from pyscf import gto
 from pyscf import scf
-from pyscf import pbc
 import trexio
 
 def to_trexio(obj, filename, backend='h5'):
@@ -46,20 +45,15 @@ def _mol_to_trexio(mol, trexio_file):
         trexio.write_nucleus_point_group(trexio_file, mol.groupname)
 
     # 2.3 Periodic boundary calculations
-    if isinstance(mol, pbc.gto.Cell):
-        # 2.3 Periodic boundary calculations (pbc group)
-        trexio.write_pbc_periodic(trexio_file, True)
+    try:
+        mol.a
+        periodic = True
+        raise NotImplementedError('PBC is not supported yet.')
         # 2.2 Cell
-        Bohr = lib.param.BOHR
-        a = np.array(mol.a[0]) / Bohr  # angstrom -> bohr
-        b = np.array(mol.a[1]) / Bohr  # angstrom -> bohr
-        c = np.array(mol.a[2]) / Bohr  # angstrom -> bohr
-        trexio.write_cell_a(trexio_file, a)
-        trexio.write_cell_b(trexio_file, b)
-        trexio.write_cell_c(trexio_file, c)
-    else:
-        # 2.3 Periodic boundary calculations (pbc group)
-        trexio.write_pbc_periodic(trexio_file, False)
+
+    except AttributeError:
+        periodic = False
+        trexio.write_pbc_periodic(trexio_file, periodic)
 
     # 2.4 Electron (electron group)
     electron_up_num, electron_dn_num = mol.nelec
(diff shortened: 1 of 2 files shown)


matthew-hennefarth:
> The ECP part works. However, a problem I found is that the `cc-pVXZ` basis sets accompanied with `ccECP`, which are often used for QMC calculations, did not pass the following assertion line in `trexio.py`.
> 
> ```
>   assert all(mol._bas[:,gto.NCTR_OF] == 1)
> ```

It seems that `mol._bas` is the internal representation for the libcint library, and `NCTR_OF` is a constant being 3 (which I believe represents number of cartesians; ie x, y, z). Unfortunately, I do not know what the internal representation of `_bas` is supposed to be.

Perhaps, since @sunqm wrote the assert statement and knows the details of libcint, he can provide you with more guidance.

kousuke-nakano:
@matthew-hennefarth, thank you very much. I will wait for @sunqm's reply.

kousuke-nakano:
@matthew-hennefarth 

I found that `mol._bas[:,gto.NCTR_OF]` refers to the numbers of generalized contractions (i.e., one set of exponents and multiple sets of coefficients, like molpro and molcas). See this [issue](https://github.com/TREX-CoE/trexio_tools/issues/42#issuecomment-2463880713). Do you know how to switch it off with some option when we build a `mol`?

matthew-hennefarth:
> I found that `mol._bas[:,gto.NCTR_OF]` refers to the numbers of generalized contractions (i.e., one set of exponents and multiple sets of coefficients, like molpro and molcas). See this [issue](https://github.com/TREX-CoE/trexio_tools/issues/42#issuecomment-2463880713). Do you know how to switch it off with some option when we build a `mol`?

I am not sure how to turn it off, sorry. From the discussion you linked it seems like the best bet would be to just manually expand the contraction set.

kousuke-nakano:
Dear @matthew-hennefarth, Sure. Thank you! Anyway, this is nothing with the ECP part. 

I have completed implementing a function to *write* the ECP part to a `TREXIO` file. The *read* function is my ToDO work. In addition, I have added several missing factors, such as `AO_normalization`, see #80.

pyscf/tools/trexio.py
```diff
@@ -27,15 +29,39 @@ def to_trexio(obj, filename, backend='h5'):
...
     trexio.write_nucleus_label(trexio_file, labels)
-    if mol._ecp:
-        raise NotImplementedError
-    trexio.write_nucleus_charge(trexio_file, mol.atom_charges())
-    trexio.write_nucleus_coord(trexio_file, mol.atom_coords())
+    if mol.symmetry:
+        trexio.write_nucleus_point_group(trexio_file, mol.groupname)
 
+    # 2.3 Periodic boundary calculations
+    try:
```
sunqm, line 48: If you don't need to consider the compatibility of old pyscf versions, PBC can be checked using isinstance(mol, pbc.gto.Cell) . Starting from pyscf-2.5, gto.Mole and pbc.gto.Cell represent two distinct classes.

kousuke-nakano: Dear @sunqm, Sure, Thank you very much. I am now implementing the PBC part, in which I do this check as you proposed.

kousuke-nakano:
Sorry @sunqm, I committed new changes ([6949296](https://github.com/pyscf/pyscf-forge/pull/68/commits/694929638f2f89892928e5f29a90ea180e39fcfe)) by mistake. I have reverted it ([db9cde9](https://github.com/pyscf/pyscf-forge/pull/68/commits/db9cde9e7c5128f5f118e939ac52d48ca4d9e812)).

sunqm (approved):

kousuke-nakano:
@sunqm, I have resolved a trivial conflict.

kousuke-nakano:
Dear @sunqm, Thank you for running the workflows. One of the workflows failed at the part unrelated to this pull request.

q-posev:
@kousuke-nakano are you done with this PR? 

I left one comment about the UHF case, let me know if you prefer to fix it here or if you prefer me to do it in another PR. It should not be too complicated, just a question of extracting the number of alpha and beta electrons from the UHF object.

kousuke-nakano:
Dear @q-posev, @sunqm has already implemented both RHF and UHF in #60.

q-posev:
@kousuke-nakano yes, but on line 63 in `trexio.py` in this PR you added the following line:
```python
    electron_up_num, electron_dn_num = mol.nelec
```
which is not true for UHF. 

The original code of @sunqm was not producing the number of electrons.

q-posev:
The electron writing can be probably moved to `_scf_to_trexio` function where we have an if/else block depending on the output of ` isinstance(mf, scf.uhf.UHF)` (I mean [this line](https://github.com/pyscf/pyscf-forge/pull/68/files#diff-ec9a510320bbc0d6f50c9c5364cfa300ee471c2770c84656d4cdf70864e8384dR206)).

kousuke-nakano:
Dear @q-posev,

> @kousuke-nakano yes, but on line 63 in `trexio.py` in this PR you added the following line:
> 
> ```python
>     electron_up_num, electron_dn_num = mol.nelec
> ```
> 
> which is not true for UHF.
> 
> The original code of @sunqm was not producing the number of electrons.

The following code gives the correct numbers of up and down electrons. 

> import pyscf
> 
> mol_S_0 = pyscf.M(atom='H 0 0 0; F 0 0 1', basis='6-31g*', cart=True)
> electron_up_num, electron_dn_num = mol_S_0.nelec
> print(electron_up_num, electron_dn_num) # -> 5 5
> 
> mol_S_2 = pyscf.M(atom='H 0 0 0; F 0 0 1', basis='6-31g*', cart=True)
> mol_S_2.spin = 2
> electron_up_num, electron_dn_num = mol_S_2.nelec
> print(electron_up_num, electron_dn_num) # -> 6 4

q-posev:
@kousuke-nakano you are totally right, my bad! Please ignore my comments above.

Issue #52, opened 2024-06-06
double hybrid functional


sunqm:
Consider to apply https://github.com/pyscf/pyscf/issues/911

MatthewRHermes:
This is closed by #196

PR #70, opened 2024-09-27
A new LibXC interface and new MC-DCFT code

Dayou-Zhang:
This pull request consists of two contributions:

1. It implements a new LibXC interface. 
The new interface uses the CFFI library to call LibXC. It is located in a separate package named `dft2`. The new interface is fully backward compatible with the current interface as demonstrated by the test cases. It is implemented based on the PySCF 2.7 LibXC interface code. Compared to the current LibXC interface, the new interface has several advantages:
  * All underlying LibXC data structure are now cached in the Python instead of in the C library. This allows direct Python access to the underlying LibXC data structure, which provides users with more flexibility in case they need access to LibXC features for which PySCF have not yet implemented an interface.
  * As a result, most code from `lib/dft/libxc_itrf.c` are not needed and can be replaced by a Python equivalent, as performed in this pull request. This makes it easier for future interface development, in case LibXC introduces new features such as double hybrid functionals as mentioned in #52.
  * It provides an interface to set external parameters of LibXC functionals. This is one of the key motivations of this pull request because many of our work involves setting functional parameters, which PySCF currently does not support this feature yet.
  * The current DFT interface uses a string to represent a density functional in the Python layer and construct LibXC data structure in C for every API call. As a result, the LibXC data structures are constructed and destructed many times during a DFT calculation, which adds on some overhead. This can be avoided by caching the LibXC data structure in Python and reusing them throughout the entire calculation. In the new interface, a class named `XCFunctional` is provided, which is intended to be served as a high-level API representing an exchange–correlation functional, and perhaps potentially be used to replace the `xc_code` string in DFT codes to enhance performance. The design of the high-level API is still not mature, so I am open to suggestions from the PySCF community.

2. It includes the code reflecting the latest development of MC-DCFT.
This pull request enhances the code for multiconfiguration density coherence functional theory (MC-DCFT, doi: [10.1021/acs.jctc.0c01346](https://doi.org/10.1021/acs.jctc.0c01346)) by integrating with the abovementioned LibXC interface. It allows users to perform calculations with our recently developed density coherence functional DC24 (doi: [10.26434/chemrxiv-2024-78tt8](https://doi.org/10.26434/chemrxiv-2024-78tt8)).


sunqm:
I took a quick look at the cmake part, and I noticed it adds a CFFI build section. Is the CFFI build necessary? Using CFFI for the interface would require building a wheel for each Python version, which adds efforts to devops maintenance.

MatthewRHermes:
1. It looks like `pyscf/dft2/libxc.py` and `pyscf/lib/dft/libxc_itrf2.c` (at least) originate from copy-pasted source files in the PySCF core modules. Could you add a comment to the top of both files identifying the file you copied, the PySCF core-module commit from which you made the copy, and describing briefly the modifications? This will make it easier to debug in the future when continued updates to the PySCF core modules will inevitably need to be reproduced.
2. I would recommend adding an example script and unittests in which you modify functional parameters using this interface (i.e., the basic reason that you had to rewrite libxc.py), other than just checking DC24.
3. I itch a little seeing a `.so` file anywhere other than the `pyscf/lib` folder. Is there a compelling reason that `libxc_cffi.*.so` can't go there?

MatthewRHermes:
> I took a quick look at the cmake part, and I noticed it adds a CFFI build section. Is the CFFI build necessary? Using CFFI for the interface would require building a wheel for each Python version, which adds efforts to devops maintenance.

If I understand correctly, the use of CFFI is basically an alternative to writing several new functions in `libxc_itrf.c', as in the final version of [this PR](https://github.com/pyscf/pyscf/pull/1746/).

Dayou-Zhang:
> I took a quick look at the cmake part, and I noticed it adds a CFFI build section. Is the CFFI build necessary? Using CFFI for the interface would require building a wheel for each Python version, which adds efforts to devops maintenance.

The current implementation is performed in the CFFI API mode, which requires a CFFI build. If we re-write in ABI mode, it will probably not require a build, but it is not the recommended way according to the [manual](https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api), as ABI mode is slower and is more prone to problems. Therefore, I personally prefer API mode.

> 1. It looks like `pyscf/dft2/libxc.py` and `pyscf/lib/dft/libxc_itrf2.c` (at least) originate from copy-pasted source files in the PySCF core modules. Could you add a comment to the top of both files identifying the file you copied, the PySCF core-module commit from which you made the copy, and describing briefly the modifications? This will make it easier to debug in the future when continued updates to the PySCF core modules will inevitably need to be reproduced.

This is a good suggestion. I added the comments in the files that are adapted from PySCF core modules.
 
>  2. I would recommend adding an example script and unittests in which you modify functional parameters using this interface (i.e., the basic reason that you had to rewrite libxc.py), other than just checking DC24.

There is already one test case in `dft2/test/test_libxc_functional.py` where B97-1 and B97-3 `XCFunctional` objects are set to the parameters of B97-2. Currently I cannot think of a useful example script to add because we have not integrated the interface with other modules (such as Kohn-Sham module) yet, and the MC-DCFT module already has a straightforward high-level API for setting functional parameters.

>  3. I itch a little seeing a `.so` file anywhere other than the `pyscf/lib` folder. Is there a compelling reason that `libxc_cffi.*.so` can't go there?

The reason I prefer to put ` libxc_cffi.*.so ` in `dft2` is because one can easily do `from .libxc_cffi import ffi as _ffi, lib as _lib` in `libxc.py`. It is not loaded using `lib.load_library` like other ctypes libraries. We can put it in `lib`, but it is less straightforward to import the library. I have seen some Python libraries (e.g., numpy and scipy) mixing pure python code and `.so` files, so I am not sure if it is preferable to put all `.so` files in `lib`.

Dayou-Zhang:
I have now implemented the interface using the CFFI ABI mode, which should no longer need a separate wheel for each Python version. This will be the default mode when users install the package with `pip install`. However, users still have the option to compile the CFFI library using API mode from source if they pass the `-DBUILD_CFFI_API_MODE=ON` option to cmake.

MatthewRHermes:
> > 2. I would recommend adding an example script and unittests in which you modify functional parameters using this interface (i.e., the basic reason that you had to rewrite libxc.py), other than just checking DC24.
> 
> There is already one test case in `dft2/test/test_libxc_functional.py` where B97-1 and B97-3 `XCFunctional` objects are set to the parameters of B97-2. Currently I cannot think of a useful example script to add because we have not integrated the interface with other modules (such as Kohn-Sham module) yet, and the MC-DCFT module already has a straightforward high-level API for setting functional parameters.

OK, but can you add an example script for the use of that straightforward high-level API?

Dayou-Zhang:
> > > 2. I would recommend adding an example script and unittests in which you modify functional parameters using this interface (i.e., the basic reason that you had to rewrite libxc.py), other than just checking DC24.
> > 
> > 
> > There is already one test case in `dft2/test/test_libxc_functional.py` where B97-1 and B97-3 `XCFunctional` objects are set to the parameters of B97-2. Currently I cannot think of a useful example script to add because we have not integrated the interface with other modules (such as Kohn-Sham module) yet, and the MC-DCFT module already has a straightforward high-level API for setting functional parameters.
> 
> OK, but can you add an example script for the use of that straightforward high-level API?

Yes, I just added an example script to show how users can set functional parameters using the high-level API in MC-DCFT.

matthew-hennefarth:
Can you also add the meta-GGA functional available for MC-PDFT as well? This way OpenMolcas (with this [PR](https://gitlab.com/Molcas/OpenMolcas/-/merge_requests/712)) and PySCF would be up to date with regards to functional functionality.

sunqm:
After carefully consideration, I strongly recommend removing the dependency on `cffi` as it introduces unnecessary complexity. While using `ctypes` may be not as elegant as `cffi`, it is sufficient to provide the functionality here. It is not a bottleneck here and it aligns better with our current architecture. A straightforward and stupid solution is better for long-term manageability.

MatthewRHermes:
> > 1. It looks like `pyscf/dft2/libxc.py` and `pyscf/lib/dft/libxc_itrf2.c` (at least) originate from copy-pasted source files in the PySCF core modules. Could you add a comment to the top of both files identifying the file you copied, the PySCF core-module commit from which you made the copy, and describing briefly the modifications? This will make it easier to debug in the future when continued updates to the PySCF core modules will inevitably need to be reproduced.
> 
> This is a good suggestion. I added the comments in the files that are adapted from PySCF core modules.

Could you describe your modifications and report the commit hash from which you copied those files in those comments?

Dayou-Zhang:
I finished implementing a new LibXC interface based on `ctypes`. The new interface is still backward compatible with the current PySCF core module. In the new interface, the LibXC C data structure are cached and accessible in Python using `ctypes`. Each functional is still identified by the `xc_code` string as in the current PySCF core module. Users can call `register_custom_functional_` to register a new user-defined functional with a user-selected `xc_code`. This function includes a convenient interface to set most functional parameters. I included an example showing how users can perform a KS-DFT calculation using the new interface with custom functionals.

In this pull request, `pyscf/lib/dft/libxc_itrf2.c`, `pyscf/dft2/test/test_libxc.py`, `pyscf/dft2/test/dm_h4.npy`, and `pyscf/dft2/libxc.py` are adapted or copied from the PySCF core module commit 25eaa9572977b903de24d5c11ad345cecd744728. I put the commit hash in the source code comments.

I also updated the MC-DCFT code accordingly to use the new LibXC interface.

pyscf/dft2/libxc.py
MatthewRHermes, line 1: Is it accurate to summarize the major differences between this file and its PySCF core-module original form as "Implemented functions register_custom_functional_, update_custom_functional_, and unregister_custom_functional_. To make this new functionality possible, interface functions which originally took arguments of type c_int were modified to take arguments of type xc_func_type from LibXC."? If that's accurate, can those sentences be added to the comment block at the top of the file? You don't have to describe every little thing, but a basic orientation will be very helpful if/when the core-module libxc.py is ever updated.

pyscf/lib/dft/libxc_itrf2.c
MatthewRHermes, line 1: Is it accurate to summarize the major differences between this file and its PySCF core-module original form as "Interface functions which originally took arguments of type int were modified to take arguments of type xc_func_type from LibXC."? If that's accurate, can that sentence be added to the comment block at the top of the file?

I notice you also added some `const` and `restrict` keywords in various places to otherwise unaltered functions, is that necessary or is it just a minor optimization? It's NBD either way.

MatthewRHermes (approved):
I think it's good, overall. The copy-pasting of large functions like `eval_xc` means we have to be vigilant about changes to the core-module versions of those functions that we may or may not want to port, but since libxc.py and itrf.c are pretty stable in the core modules so it shouldn't be a huge burden, and on the upside it makes it easier if the core-module maintainers ever decide that they want this functionality.

pyscf/dft2/libxc.py
```diff
@@ -0,0 +1,1231 @@
...
+_DOUBLE_ARRAY_TYPE = _ffi.typeof("double[]")
+_DOUBLE_ARRAY_2_TYPE = _ffi.typeof("double[2]")
+_DOUBLE_ARRAY_3_TYPE = _ffi.typeof("double[3]")
+
+def xc_func_init(xc_code, nspin=_lib.XC_UNPOLARIZED):
+    func = _ffi.new(_XC_FUNC_TYPE_PTR_TYPE)
+    if _lib.xc_func_init(func, xc_code, nspin) != 0:
+        raise ValueError(f"Functional '{xc_code}' not found")
+    #  xc_func_end() will automatically be called when `func` is destructed
+    return _ffi.gc(func, _lib.xc_func_end)
```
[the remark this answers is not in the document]
Dayou-Zhang: In my implementation, I used void* to pass the address of structs to LibXC. This is because LibXC already has many APIs to get and set functional properties. Therefore, I think exposing the members of the structs in Python is not really necessary. Only the address of the struct is important, and that is what void* does. If users really need access to the struct members that are not exposed by the API, they may either write a C library or explicitly declare the struct members in ctypes.

pyscf/lib/dft/libxc_itrf2.c
```diff
@@ -0,0 +1,1147 @@
...
+ * XC_FAMILY_OEP          16
+ * XC_FAMILY_HYB_GGA      32
+ * XC_FAMILY_HYB_MGGA     64
+ * XC_FAMILY_HYB_LDA     128
+ */
+int LIBXC_xc_type(const int nfunc, const xc_func_type *func)
+{
+        int i;
+        int type = -1;
+        for (i = 0; i < nfunc; i++) {
```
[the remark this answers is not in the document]
Dayou-Zhang: This is because in the new implementation, functionals are processed in units of each `xc_code`, rather than in each integer ID. It would be clearer (and might be slightly more efficient) to process all the functional components of each `xc_code` at once in the C layer whenever possible.

pyscf/dft2/libxc.py
Dayou-Zhang: Another motivation of the new interface is to improve the memory management of the LibXC data structure, which would not need to be constructed and destructed at every API call. I added the following paragraph to the comment block of `pyscf/dft2/libxc.py`:
```
# Implemented functions register_custom_functional_, update_custom_functional_,
# and unregister_custom_functional_. To make this new functionality possible,
# and to improve the memory management of the LibXC data structure, interface
# functions which originally took arguments of type c_int were modified to take
# arguments of type xc_func_type from LibXC. The LibXC data structure are cached
# and managed in the python layer to avoid construction and destruction at every
# API call.
```

pyscf/lib/dft/libxc_itrf2.c
Dayou-Zhang: This is just a minor optimization. Using `restrict` tells the compiler that two arrays will not overlap. Providing extra hints can be very helpful for the compiler to produce more efficient code.

MatthewRHermes: Okay, but premature optimization is the root of many kinds of evil. `const` is pretty harmless, but `restrict` scares me a little. But I'm testing this on many other things and not seeing any problems, so it's probably fine.

Issue #79, opened 2024-11-09
CASCI/CASSCF for TREXIO


NastaMauger:
Hello,

Regarding this PR #60, I realized that it is currently not possible to export objects like CASCI/CASSCF object to TREXIO. It would also be helpful to have the ability to register these objects in TREXIO. I'd be interested in contributing, but I was unable to extract the necessary information from get_h1eff()[0] (see this thread: [TREX-CoE/trexio_tools Issue #46](https://github.com/TREX-CoE/trexio_tools/issues/46#issue-2646327032)).

Best

sunqm:
The second object returned by `casci.get_h1eff()` is the energy contribution from the core determinant. I didn't see that in your issue post. Is that part missed?

NastaMauger:
Thank your for your reply @sunqm 

Your answer confuses me because I couldn't find anything about registering get_h1eff()[1] in the TREXIO documentation.

Also, can the `write_eri `function be used to write `h2_eff`, or would modifications be needed due to the effective space? I noticed that this function is registered in the MO field of TREXIO, whereas `get_h2cas` seems to be in the AO field, isn't it?

MatthewRHermes:
`get_h2eff` and `get_h2cas` are exactly the same thing nowadays.

PR #80, opened 2024-11-10
The User can now register its determinant into trexio format

NastaMauger:
Hello,

For a specific reason, I needed to register my determinants into TREXIO, so I modified the code to accomplish this.

Since the functions use the MCSCF object and `mcscf_to_trexio` has not been developed yet, it might be beneficial to register the determinants infos into a different TREXIO file compared to the one used for the mf part. Therefore, this final TREXIO file only contains:

- Number of MOs
- Number of alpha electrons
- Number of beta electrons
- Total number of electrons
- Number of determinants
- Coefficients
- Determinants in bitfield format

The user can also set a threshold to register into the TREXIO file only the determinants with coefficients higher than the user-defined value. If no value is provided by the user, the full list of determinants generated by the PySCF object is registered.

As an improvement, one could develop a `mcscf_to_trexio function` (in progress [here](https://github.com/pyscf/pyscf-forge/issues/79#issue-2646546761)) as well as a converter from bitfield to a human-readable format. This would allow users to import a TREXIO file from other software and read it easily.

The different functions from this PR can be called as follow:
```
norb = 10
nelec = 4
ci_threshold = 1e-5
mcscf.kernel()
det_to_trexio(mcscf, norb, nelec, 'my_beautiful_determiants.hdf5', ci_threshold = ci_threshold)
num_det, coeff, det = read_det_trexio('my_beautiful_determiants.hdf5')

print(num_det)
print(coeff)
print(det) #In Bitfiled format
```

Best

Edit: As recommended [here](https://github.com/pyscf/pyscf-forge/pull/80#issuecomment-2468766145),  two files are attached where all new functionalities are working on CASCI/sCI and FCI objects.
[casci.txt](https://github.com/user-attachments/files/17736768/casci.txt)
[fci.txt](https://github.com/user-attachments/files/17736771/fci.txt)

### DIFF
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -252,9 +252,10 @@ def get_occsa_and_occsb(mcscf, norb, nelec):
 
     return occsa_sorted, occsb_sorted, ci_values_sorted, num_determinants
 
-def det_to_trexio(mcscf, norb, nelec, mo_num, filename, backend='h5', chunk_size=100000):  
+def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', chunk_size=100000):  
     from trexio_tools.group_tools import determinant as trexio_det
 
+    mo_num = norb
     int64_num = int((mo_num - 1) / 64) + 1 
     occsa, occsb, ci_values, num_determinants = get_occsa_and_occsb(mcscf, norb, nelec)
 
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -299,6 +299,5 @@ def read_det_trexio(filename):
         num_det = trexio.read_determinant_num(tf)
         coeff = trexio.read_determinant_coefficient(tf, offset_file, num_det)
         det = trexio.read_determinant_list(tf, offset_file, num_det)
-  
         return num_det, coeff, det
 
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -273,6 +273,9 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5'):
       if(trexio.has_determinant(tf)):
         trexio.delete_determinant(tf)
       trexio.write_mo_num(tf, mo_num)
+      trexio.write_electron_up_num(tf, len(a))
+      trexio.write_electron_dn_num(tf, len(b))
+      trexio.write_electron_num(tf, len(a)+len(b))
       trexio.write_determinant_list(tf, offset_file, num_determinants, det_list)
       trexio.write_determinant_coefficient(tf, offset_file, num_determinants, ci_values)
 
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -16,7 +16,7 @@ import math
 from pyscf import lib
 from pyscf import gto
 from pyscf import scf
-from pyscf import fci 
+from pyscf import fci
 import trexio
 
 def to_trexio(obj, filename, backend='h5'):
@@ -133,7 +133,7 @@ def mol_from_trexio(filename):
     return mol.build()
 
 def scf_from_trexio(filename):
-    mol = mol_from_trexio(filename, backend)
+    mol = mol_from_trexio(filename)
     with trexio.File(filename, 'r', back_end=trexio.TREXIO_AUTO) as tf:
         mo_energy = trexio.read_mo_energy(tf)
         mo        = trexio.read_mo_coefficient(tf)
@@ -256,7 +256,7 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., c
     from trexio_tools.group_tools import determinant as trexio_det
 
     mo_num = norb
-    int64_num = int((mo_num - 1) / 64) + 1 
+    int64_num = int((mo_num - 1) / 64) + 1
     occsa, occsb, ci_values, num_determinants = get_occsa_and_occsb(mcscf, norb, nelec, ci_threshold)
 
     det_list = []
@@ -271,9 +271,9 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., c
     if num_determinants > chunk_size:
         n_chunks = math.ceil(num_determinants / chunk_size)
     else:
-        n_chunks = 1 
+        n_chunks = 1
 
-    with trexio.File(filename, 'u', back_end=_mode(backend)) as tf: 
+    with trexio.File(filename, 'u', back_end=_mode(backend)) as tf:
         if trexio.has_determinant(tf):
             trexio.delete_determinant(tf)
         trexio.write_mo_num(tf, mo_num)
@@ -281,7 +281,7 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., c
         trexio.write_electron_dn_num(tf, len(b))
         trexio.write_electron_num(tf, len(a) + len(b))
 
-        offset_file = 0 
+        offset_file = 0
         for i in range(n_chunks):
             start = i * chunk_size
             end = min((i + 1) * chunk_size, num_determinants)
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -260,31 +260,33 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5'):
 
     det_list = []
     for a, b, coeff in zip(occsa, occsb, ci_values):
-      det_tmp     = []
-      det_tmp    += trexio_det.to_determinant_list(occsa, int64_num)
-      det_tmp    += trexio_det.to_determinant_list(occsb, int64_num)
-      det_list.append(det_tmp)
+        occsa_upshifted = [orb + 1 for orb in a]  
+        occsb_upshifted = [orb + 1 for orb in b]
+        det_tmp     = []
+        det_tmp    += trexio_det.to_determinant_list(occsa_upshifted, int64_num)
+        det_tmp    += trexio_det.to_determinant_list(occsb_upshifted, int64_num)
+        det_list.append(det_tmp)
     
     offset_file = 0
 
     with trexio.File(filename, 'u', back_end=_mode(backend)) as tf:
-      if(trexio.has_determinant(tf)):
-        trexio.delete_determinant(tf)
-      trexio.write_mo_num(tf, mo_num)
-      trexio.write_electron_up_num(tf, len(a))
-      trexio.write_electron_dn_num(tf, len(b))
-      trexio.write_electron_num(tf, len(a)+len(b))
-      trexio.write_determinant_list(tf, offset_file, num_determinants, det_list)
-      trexio.write_determinant_coefficient(tf, offset_file, num_determinants, ci_values)
+        if(trexio.has_determinant(tf)):
+            trexio.delete_determinant(tf)
+        trexio.write_mo_num(tf, mo_num)
+        trexio.write_electron_up_num(tf, len(a))
+        trexio.write_electron_dn_num(tf, len(b))
+        trexio.write_electron_num(tf, len(a)+len(b))
+        trexio.write_determinant_list(tf, offset_file, num_determinants, det_list)
+        trexio.write_determinant_coefficient(tf, offset_file, num_determinants, ci_values)
 
 
 def read_det_trexio(filename, backend='h5'):
     with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
-      offset_file = 0
+        offset_file = 0
 
-      num_det = trexio.read_determinant_num(tf)
-      coeff = trexio.read_determinant_coefficient(tf, offset_file, num_det)
-      det = trexio.read_determinant_list(tf, offset_file, num_det)
+        num_det = trexio.read_determinant_num(tf)
+        coeff = trexio.read_determinant_coefficient(tf, offset_file, num_det)
+        det = trexio.read_determinant_list(tf, offset_file, num_det)
   
-      return num_det, coeff[0], det[0]
+        return num_det, coeff[0], det[0]
 
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -93,9 +93,10 @@ def _cc_to_trexio(cc_obj, trexio_file):
 def _mcscf_to_trexio(cas_obj, trexio_file):
     raise NotImplementedError
 
-def mol_from_trexio(filename,backend=trexio.TREXIO_AUTO):
+
+def mol_from_trexio(filename, backend='h5'):
     mol = gto.Mole()
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
         assert trexio.read_basis_type(tf) == 'Gaussian'
         if trexio.has_ecp(tf):
             raise NotImplementedError
@@ -132,9 +133,9 @@ def mol_from_trexio(filename,backend=trexio.TREXIO_AUTO):
     mol._basis = basis
     return mol.build()
 
-def scf_from_trexio(filename, backend=trexio.TREXIO_AUTO):
+def scf_from_trexio(filename, backend='h5'):
     mol = mol_from_trexio(filename, backend)
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
         mo_energy = trexio.read_mo_energy(tf)
         mo        = trexio.read_mo_coefficient(tf)
         mo_occ    = trexio.read_mo_occupation(tf)
@@ -174,9 +175,9 @@ def write_eri(eri, filename, backend='h5'):
     with trexio.File(filename, 'w', back_end=_mode(backend)) as tf:
         trexio.write_mo_2e_int_eri(tf, 0, num_integrals, idx, eri.ravel())
 
-def read_eri(filename, backend=trexio.TREXIO_AUTO):
+def read_eri(filename, backend='h5'):
     '''Read ERIs in AO basis, 8-fold symmetry is assumed'''
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
         nmo = trexio.read_mo_num(tf)
         nao_pair = nmo * (nmo+1) // 2
         eri_size = nao_pair * (nao_pair+1) // 2
@@ -280,7 +281,7 @@ def det_to_trexio(mcscf, norb, nelec, ci_threshold_or_filename, filename=None, b
     else:
         n_chunks = 1 
 
-    with trexio.File(filename, 'u', back_end=_mode(backend)) as tf: 
+    with trexio.File(filename, 'w', back_end=_mode(backend)) as tf: 
         if trexio.has_determinant(tf):
             trexio.delete_determinant(tf)
         trexio.write_mo_num(tf, mo_num)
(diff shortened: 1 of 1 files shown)


NastaMauger:
Following discussion from https://github.com/TREX-CoE/trexio_tools/issues/45#issuecomment-2466704935

pyscf/tools/trexio.py
```diff
@@ -260,8 +260,6 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5'):
 
     det_list = []
     for a, b, coeff in zip(occsa, occsb, ci_values):
-      occsa_upshifted = [orb + 1 for orb in a]  
-      occsb_upshifted = [orb + 1 for orb in b]
       det_tmp     = []
       det_tmp    += trexio_det.to_determinant_list(occsa_upshifted, int64_num)
```
q-posev, line 264: @NastaMauger you forgot to rename `occsa_upshifted` to `occsa` here (and same for `occsb`).

NastaMauger: My bad ! It is done now

sunqm:
The code mixes 2-space and 4-space indention. Please use consistent indentions.

@q-posev I'm not very familiar with TrexIO. I'd appreciate your feedback on this new feature. Please let me know if you think this code is ready for merging. Thanks!

NastaMauger:
@q-posev I realized that the original code (which I have not modified) do not register the ao shell and atomic normalization.
It does not prevent the pyscf-forge to work but this missing infos into trexio file might be an issue for some software (such as qp2).
I was not able to extract this infos from original PySCF mol object

pyscf/tools/trexio.py
```diff
@@ -223,3 +224,69 @@ def _group_by(a, keys):
...
+    selected_occslst = occslst[:num_determinants]
+
+    occsa = []
+    occsb = []
+    ci_values = []
+
+    for i in range(min(len(selected_occslst), mcscf.ci.shape[0])):
+        for j in range(min(len(selected_occslst), mcscf.ci.shape[1])):
+            ci_coeff = mcscf.ci[i, j]
+            if np.abs(ci_coeff) > 1e-8:  # Check if CI coefficient is significant
```
q-posev, line 241: Can we make this threshold for CI coefficients configurable as an argument to this function? I would trust the user to have the full control of the determinant filtering (we can keep your 1e-8 as default).

pyscf/tools/trexio.py
```diff
@@ -223,3 +224,69 @@ def _group_by(a, keys):
...
+        trexio.write_electron_num(tf, len(a)+len(b))
+        trexio.write_determinant_list(tf, offset_file, num_determinants, det_list)
+        trexio.write_determinant_coefficient(tf, offset_file, num_determinants, ci_values)
+
+
+def read_det_trexio(filename, backend='h5'):
+    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+        offset_file = 0
+
+        num_det = trexio.read_determinant_num(tf)
```
q-posev, line 287: This is more of a suggestion (I do not insist for it to be implemented here): If one day `num_det` becomes huge - the user can run out of memory while reading the CI info. This is why we allow to read/write determinants in buffers (or chunks) of fixed size. It might be a good idea to define some relatively big buffer_size (e.g. 100k) and if `num_det` is larger than that value - read the data in chunks. Example of writing in chunks (using `offset_file` to advance the data pointers) can be found in [trexio-tutorials](https://github.com/TREX-CoE/trexio-tutorials/blob/master/notebooks/how_to_determinants.md).

q-posev:
@NastaMauger I left a few comments but overall the code looks great, thank you for your contribution! Does your code work for both selected CI and FCI objects of PySCF?

@sunqm Sure, with pleasure! I left a few comments on this PR. I would like to also take this opportunity to give you kudos for all the hard work that you and the team do on PySCF - it's a great software!

q-posev:
@sunqm if I may: we have been recently discussing a few issues that TREXIO users I can summarize them as follows:
1. Failures to produce TREXIO files from PySCF when using Dunning cc-pvXz basis sets due to the contraction scheme being used, reported [here](https://github.com/TREX-CoE/trexio_tools/issues/42) by @scemama , @NastaMauger and @kousuke-nakano . @scemama proposed a solution [here](https://github.com/TREX-CoE/trexio_tools/issues/42#issuecomment-2464263279) which would basically "decontract"/expand the basis set. Do you think it is feasible? Are there any internal PySCF objects which hold the necessary pieces?
2. As mentioned by @NastaMauger [here](https://github.com/pyscf/pyscf-forge/pull/60): the current TREXIO interface implemented in `pyscf-forge` PR #60 does not produce some details about the atomic orbitals (we are only missing `shell` and their `normalization` arrays from the [ao group in TREXIO](https://trex-coe.github.io/trexio/trex.html#orgf00657e) format), which prevents importing of the PySCF-produced TREXIO files in some other codes. Would it be possible to add this information to the TREXIO file produced by the `_mol_to_trexio` function?

pyscf/tools/trexio.py
NastaMauger: @q-posev I am not sure if setting a CI threshold is a good idea. For example, after a sCI run, my system produces the following (full) list of determinants with CI values using a threshold of 1e-8:
```
occsa    occsb       CI coeff
[0 1]       [0 1]    -0.99779049148461
[0 2]       [0 2]     0.03340264667473
[0 2]       [1 2]    -0.03321881541135
[1 2]       [0 2]    -0.03321881541135
[1 2]       [1 2]     0.03303662972152
```

Now, let’s say a user wants to keep only values higher than 0.1. In this case, the code would register only [0 1] [0 1] -0.9977904914846 and set the others to 0. However, this corresponds to truncating the wavefunctions, which is a different process, right? This would also imply that the one- and two-electron integrals would need to be renormalized to account for the new threshold.

I chose a threshold of 1e-8, as I believe it’s sufficient to filter out coefficients that are effectively zero in terms of bit precision.

Let me know your thoughts.

pyscf/tools/trexio.py
```diff
@@ -223,3 +224,69 @@ def _group_by(a, keys):
...
+        trexio.write_mo_num(tf, mo_num)
+        trexio.write_electron_up_num(tf, len(a))
+        trexio.write_electron_dn_num(tf, len(b))
+        trexio.write_electron_num(tf, len(a)+len(b))
+        trexio.write_determinant_list(tf, offset_file, num_determinants, det_list)
+        trexio.write_determinant_coefficient(tf, offset_file, num_determinants, ci_values)
+
+
+def read_det_trexio(filename, backend='h5'):
+    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
```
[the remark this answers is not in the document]
NastaMauger: I set `back_end=_mode(backend)` to maintain consistency with how it was previously written in the code. For example, the same back_end is called on line 95.

I can certainly change it to` back_end=trexio.TREXIO_AUTO` as you advised, but I thought it might be better to stay consistent.

We could either keep it as it is or update all calls to use `back_end=trexio.TREXIO_AUTO`. Let me know which option would be best.

pyscf/tools/trexio.py
NastaMauger: Done

kousuke-nakano:
@sunqm and @q-posev I have already implemented the second point in my forked pyscf-forge branch, i.e., the AO normalization. I will make a pull request soon. I have not found a solution for the first point.

pyscf/tools/trexio.py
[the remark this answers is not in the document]
q-posev: Yes, I would advise to use that but **only** for reading TREXIO files. The reason is that when one wants to write/create TREXIO file - the I/O back end has to be provided explicitly, but for reading - our library is smart enough and can detect the I/O back end that was used to produce the TREXIO file. So using `TREXIO_AUTO` allows you to stay compatible with whatever new back end we will decide to implement in the future.

If @sunqm is OK with that, I would recommend to modify also the line 95 accordingly.

Thanks!

pyscf/tools/trexio.py
NastaMauger: Done.
I decided to place the CI threshold argument right after the user-defined `nelec` value for consistency. I thought it might be counterintuitive for the user to place this threshold after the `trexio_file` argument.

Let me know your thoughts.

pyscf/tools/trexio.py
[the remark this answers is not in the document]
NastaMauger: I made this modification in all functions that use trexio.read. It’s done in a separate commit so that if @sunqm  disagree, I can easily roll it back.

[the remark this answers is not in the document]
q-posev: @NastaMauger Sorry, i should have precised that: you need to use back_end=trexio.TREXIO_AUTO in the calls to trexio.File with 'r' mode. And then you can omit the backend argument in the functions with *_from_trexio and read_eri. I am not sure your commit works as expected because _mode function doesn't handle the TREXIO_AUTO case.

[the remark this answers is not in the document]
NastaMauger: @q-posev  My apologies—I am a bit confused about the backend. I have modified the code based on your last comment. Could you please check it?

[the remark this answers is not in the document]
sunqm: > Yes, I would advise to use that but **only** for reading TREXIO files. The reason is that when one wants to write/create TREXIO file - the I/O back end has to be provided explicitly, but for reading - our library is smart enough and can detect the I/O back end that was used to produce the TREXIO file. So using `TREXIO_AUTO` allows you to stay compatible with whatever new back end we will decide to implement in the future.
> 
> If @sunqm is OK with that, I would recommend to modify also the line 95 accordingly.
> 
> Thanks!

sure. Please update the module as you like. I implemented the version based on the paper or the documents. Many places are likely sub-optimal.

sunqm:
> @sunqm if I may: we have been recently discussing a few issues reported by the TREXIO users. I can summarize them as follows:
> 
> 1. Failures to produce TREXIO files from PySCF when using Dunning cc-pvXz basis sets due to the contraction scheme being used, reported [here](https://github.com/TREX-CoE/trexio_tools/issues/42) by @scemama , @NastaMauger and @kousuke-nakano . @scemama proposed a solution [here](https://github.com/TREX-CoE/trexio_tools/issues/42#issuecomment-2464263279) which would basically "decontract"/expand the basis set. Do you think it is feasible? Are there any internal PySCF objects which hold the necessary pieces?
> 2. As mentioned by @NastaMauger [here](https://github.com/pyscf/pyscf-forge/pull/80#issuecomment-2466983271): the current TREXIO interface implemented in `pyscf-forge` PR [Trexio interface #60](https://github.com/pyscf/pyscf-forge/pull/60) does not produce some details about the atomic orbitals (we are only missing `shell` and their `normalization` arrays from the [ao group in TREXIO](https://trex-coe.github.io/trexio/trex.html#orgf00657e) format), which prevents importing of the PySCF-produced TREXIO files in some other codes. Would it be possible to add this information to the TREXIO file produced by the `_mol_to_trexio` function?

1. Is it an issue the mixing of segment contraction and general contraction in pyscf? Does basis sets like ANO work using this the interface? The cc-pv*Z basis in pyscf was obtained as an "optimized" format, which contains from basis set exchange. The zeros in the contraction coefficients are squeezed, leading to , e.g. multiple S-shell blocks, P-shell blocks. Does it work if I aggregate all S shells into one block, and P into one block, so on so forth?

2. Sure. I'm not clear how to build the index and normalization. An example could explain easily I guess. What do these variables look like for STO-3G of Carbon?

kousuke-nakano:
Dear @sunqm,

> Sure. I'm not clear how to build the index and normalization. An example could explain easily I guess. What do these variables look like for STO-3G of Carbon?

I have already implemented it in #68.

kousuke-nakano:
Dear @sunqm 

> Does basis sets like ANO work using this the interface?

No, I do not think the format of the ANO basis set is compatible with the TREXIO format.

>F    S
 103109.46                   0.0000637             -0.0000148              0.0000151             -0.0000154
  15281.007                  0.0005028             -0.0001172              0.0001187             -0.0001160
   3441.5392                 0.0026677             -0.0006239              0.0006363             -0.0006596
    967.09483                0.0112003             -0.0026280              0.0026599             -0.0025500
    314.03534                0.0390980             -0.0093503              0.0096068             -0.0102361
    113.44230                0.1122657             -0.0278490              0.0284336             -0.0267629
     44.644727               0.2472042             -0.0676880              0.0719831             -0.0829345
     18.942874               0.3680345             -0.1230542              0.1310393             -0.1141649
      8.5327430              0.2908617             -0.1522180              0.2031617             -0.3160866
      3.9194010              0.0781024             -0.0075794             -0.0915990              0.4487239
      1.5681570              0.0035403              0.3759313             -1.0070250              1.1605351
      0.6232900              0.0009157              0.5438482              0.0174627             -1.3695700
      0.2408610              0.0000765              0.2127746              0.7490027             -0.6041401
      0.0843010              0.0000205              0.0066422              0.3447692              1.2967102
 
Here, the problem is that one exponent is associated with multiple contraction coefficients,
which `TREXIO` does not expect.

Instad, for instance, the **ccecp**-cc-pv*Z basis set implemented in `pyscf` works with this interface.

> F S
      84.048652  -0.00000280
      41.094617  -0.00035430
      20.092738  0.01840880
       9.824112  -0.12776610
       4.803386  -0.07028720
       2.348560  0.15716040
       1.148301  0.37690120
       0.561449  0.40951650
       0.274514  0.19708930
       0.134220  0.02434820

pyscf/tools/trexio.py
```diff
@@ -91,9 +93,10 @@ def _cc_to_trexio(cc_obj, trexio_file):
 def _mcscf_to_trexio(cas_obj, trexio_file):
     raise NotImplementedError
 
+
 def mol_from_trexio(filename, backend='h5'):
```
q-posev, line 97: `backend='h5'` can be removed here

pyscf/tools/trexio.py
```diff
@@ -91,9 +93,10 @@ def _cc_to_trexio(cc_obj, trexio_file):
...
     mol = gto.Mole()
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
```
q-posev, line 99: `back_end` instead of `backend` (tests failed in the CI because of that)

pyscf/tools/trexio.py
```diff
@@ -132,7 +135,7 @@ def mol_from_trexio(filename, backend='h5'):
 
 def scf_from_trexio(filename, backend='h5'):
```
q-posev, line 136: same as above (`backend` option can be removed)

pyscf/tools/trexio.py
```diff
@@ -174,7 +177,7 @@ def write_eri(eri, filename, backend='h5'):
 
 def read_eri(filename, backend='h5'):
```
q-posev, line 178: same as above (`backend` option can be removed)

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+        det_tmp += trexio_det.to_determinant_list(occsa_upshifted, int64_num)
+        det_tmp += trexio_det.to_determinant_list(occsb_upshifted, int64_num)
+        det_list.append(det_tmp)
+
+    if num_determinants > chunk_size:
+        n_chunks = math.ceil(num_determinants / chunk_size)
+    else:
+        n_chunks = 1 
+
+    with trexio.File(filename, 'w', back_end=_mode(backend)) as tf:
```
q-posev, line 284: you need unsafe `'u'` mode to delete things from TREXIO file (e.g. `determinant` group as here)

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+            end = min((i + 1) * chunk_size, num_determinants)
+            current_chunk_size = end - start
+
+            if current_chunk_size > 0:
+                trexio.write_determinant_list(tf, offset_file, current_chunk_size, det_list[start:end])
+                trexio.write_determinant_coefficient(tf, offset_file, current_chunk_size, ci_values[start:end])
+                offset_file += current_chunk_size
+
+
+def read_det_trexio(filename, backend='h5'):
```
q-posev, line 304: same as above (`backend` option can be removed)

pyscf/tools/trexio.py
```diff
@@ -132,7 +135,7 @@ def mol_from_trexio(filename, backend='h5'):
 
 def scf_from_trexio(filename, backend='h5'):
     mol = mol_from_trexio(filename, backend)
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
```
q-posev, line 138: `back_end` instead of `backend`

pyscf/tools/trexio.py
```diff
@@ -174,7 +177,7 @@ def write_eri(eri, filename, backend='h5'):
 
 def read_eri(filename, backend='h5'):
     '''Read ERIs in AO basis, 8-fold symmetry is assumed'''
-    with trexio.File(filename, 'r', back_end=_mode(backend)) as tf:
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
```
q-posev, line 180: `back_end` instead of `backend`

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+            current_chunk_size = end - start
+
+            if current_chunk_size > 0:
+                trexio.write_determinant_list(tf, offset_file, current_chunk_size, det_list[start:end])
+                trexio.write_determinant_coefficient(tf, offset_file, current_chunk_size, ci_values[start:end])
+                offset_file += current_chunk_size
+
+
+def read_det_trexio(filename, backend='h5'):
+    with trexio.File(filename, 'r', backend=trexio.TREXIO_AUTO) as tf:
```
q-posev, line 305: `back_end` instead of `backend`

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+        offset_file = 0
+
+        num_det = trexio.read_determinant_num(tf)
+        coeff = trexio.read_determinant_coefficient(tf, offset_file, num_det)
+        det = trexio.read_determinant_list(tf, offset_file, num_det)
+  
+        return num_det, coeff[0], det[0]
```
q-posev, line 312: why do you return `coeff[0]` and `det[0]` here and not the full arrays?

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+
+    # Sort by the absolute value of the CI coefficients in descending order
+    sorted_indices = np.argsort(-np.abs(ci_values))
+    occsa_sorted = [occsa[idx] for idx in sorted_indices]
+    occsb_sorted = [occsb[idx] for idx in sorted_indices]
+    ci_values_sorted = [ci_values[idx] for idx in sorted_indices]
+
+    return occsa_sorted, occsb_sorted, ci_values_sorted, num_determinants
+
+def det_to_trexio(mcscf, norb, nelec, ci_threshold_or_filename, filename=None, backend='h5', chunk_size=100000):
```
q-posev, line 256: I see what you mean about the `ci_threshold` position but I think having an argument with dual meaning like here can lead to errors (dynamic typing of Python does it often).

We can keep `filename` as mandatory argument (as you did originally) and place optional `ci_threshold` before the chunk size, so we would get 
```python
def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., chunk_size=100000):
```

pyscf/tools/trexio.py
```diff
@@ -223,3 +226,88 @@ def _group_by(a, keys):
...
+    from trexio_tools.group_tools import determinant as trexio_det
+
+    # Determine if ci_threshold is provided
```
q-posev, line 259: This if/else block can be removed if we rollback to mandatory `filename` and option `ci_threshold`

q-posev (changes requested):
@NastaMauger I left a few comments to facilitate the code modifications. Right now the CI tests fail because of a typo (`trexio.File` class does not support `backend` but we use `back_end` instead). 
Sorry for the trouble with the modifications, I would push the necessary changes myself to make things easier but this PR is from your fork and I cannot push there.

pyscf/tools/trexio.py
q-posev: We now use `TREXIO_AUTO` in the reading functions, that's why the `backend` option does not need to be provided here

q-posev:
@NastaMauger thank you, it's perfect now!

@sunqm I think we are ready to merge this if the CI is green.

sunqm (approved):

PR #81, opened 2024-11-11
Parity violating contributions to molecular energy.

juanjoaucar:


pyscf/pv/test/test_energy-pv.py
```diff
@@ -0,0 +1,70 @@
...
+    mf.kernel()
+
+def tearDownModule():
+    global mf
+    mol.stdout.close()
+    del mol, mf
+
+class KnownValues(unittest.TestCase):
+    def test_pv(self):
+#       Reference results (for PV) are obtained from DIRAC24
```
MatthewRHermes, line 59: I am not very familiar with this enterprise, but is -2.745821e-21 sufficiently far from zero that we can be sure that finite-precision rounding errors and SCF tolerances are not controlling it? Is there a way to generate a fantasy system with a much larger parity violation, perhaps by manipulating fundamental constants such as the speed of light?

sunqm:
From PR https://github.com/pyscf/pyscf/pull/2459

pyscf/pv/test/test_energy-pv.py
sunqm: Maybe test a relative value?

juanjoaucar: Thanks for the suggestions.
The problem was not the finite-precision rounding errors and SCF tolerances but naive mistakes I've made while changing name variables. It should finish well now.
The tiny results are in part due to the Fermi Constant factor. They are expected to be like that.

sunqm (approved):

PR #84, opened 2024-11-20
Support contracted Gaussians in trexio interface (fix #82)

### LINKED ISSUE #82: Issues producing TREXIO file with generally contracted basis sets
> @sunqm if I may: we have been recently discussing a few issues reported by the TREXIO users. I can summarize them as follows:
> 
> 1. Failures to produce TREXIO files from PySCF when using Dunning cc-pvXz basis sets due to the contraction scheme being used, reported [here](https://github.com/TREX-CoE/trexio_tools/issues/42) by @scemama , @NastaMauger and @kousuke-nakano . @scemama proposed a solution [here](https://github.com/TREX-CoE/trexio_tools/issues/42#issuecomment-2464263279) which would basically "decontract"/expand the basis set. Do you think it is feasible? Are there any internal PySCF objects which hold the necessary pieces?
> 2. As mentioned by @NastaMauger [here](https://github.com/pyscf/pyscf-forge/pull/80#issuecomment-2466983271): the current TREXIO interface implemented in `pyscf-forge` PR [Trexio interface #60](https://github.com/pyscf/pyscf-forge/pull/60) does not produce some details about the atomic orbitals (we are only missing `shell` and their `normalization` arrays from the [ao group in TREXIO](https://trex-coe.github.io/trexio/trex.html#orgf00657e) format), which prevents importing of the PySCF-produced TREXIO files in some other codes. Would it be possible to add this information to the TREXIO file produced by the `_mol_to_trexio` function?

1. Is it an issue the mixing of segment contraction and general contraction in pyscf? Does basis sets like ANO work using this the interface? The cc-pv*Z basis in pyscf was obtained as an "optimized" format, which contains from basis set exchange. The zeros in the contraction coefficients are squeezed, leading to , e.g. multiple S-shell blocks, P-shell blocks. Does it work if I aggregate all S shells into one block, and P into one block, so on so forth?

2. Sure. I'm not clear how to build the index and normalization. An example could explain easily I guess. What do these variables look like for STO-3G of Carbon?

_Originally posted by @sunqm in https://github.com/pyscf/pyscf-forge/issues/80#issuecomment-2472347698_

sunqm:


pyscf/tools/trexio.py
```diff
@@ -438,3 +439,23 @@ def read_det_trexio(filename):
...
+        nctr = shell[gto.NCTR_OF]
+        if nctr == 1:
+            _bas.append(shell)
+            continue
+
+        nprim = shell[gto.NPRIM_OF]
+        pcoeff = shell[gto.PTR_COEFF]
+        bs = np.repeat(shell[np.newaxis], nctr, axis=0)
+        bs[:,gto.NCTR_OF] = 1
+        bs[:,gto.PTR_COEFF] = np.arange(pcoeff, pcoeff+nprim*nctr, nprim)
```
MatthewRHermes, line 456: Since this is a pointer into `_env`, shouldn't one also do something to `_env` to make this work?

sunqm: The contraction coefficients are stored in a way that the general contraction and segment contraction are compatible. It only requires an additional pointer to reference the individual shell within the contracted functions.

MatthewRHermes (approved):

PR #85, opened 2024-11-22
Implemented PBC for Gamma point (real wavefunction)

kousuke-nakano:
I have implemented PBC for Gamma point (i.e., real wavefunction) in the `trexio` converter.
An example is:

```
# load python packages
import numpy as np
from pyscf.pbc import gto, dft
from pyscf.tools import trexio

# cell info
# cell
cell = gto.M(
    verbose = 4,
    a = np.eye(3)*3.5668,
    atom = '''C     0.      0.      0.    
              C     0.8917  0.8917  0.8917
              C     1.7834  1.7834  0.    
              C     2.6751  2.6751  0.8917
              C     1.7834  0.      1.7834
              C     2.6751  0.8917  2.6751
              C     0.      1.7834  1.7834
              C     0.8917  2.6751  2.6751'''
)
cell.output = pyscf_output
cell.charge = 0
cell.spin = 0
cell.symmetry = False

# basis set
cell.basis = 'ccecp-ccpvdz'
cell.exp_to_discard = 0.2

# define ecp
cell.ecp = 'ccecp'

# cell build
cell.build(cart=False)

# DFT setting
kpt = [0.0, 0.0, 0.0]
mf = dft.rks.RKS(cell, kpt=kpt)
mf.xc = "lda_x,lda_c_pz"
mf.chkfile = "pyscf_diamond.chk"

# run HF/DFT calc.
total_energy = mf.kernel()
print(f"Total HF/DFT energy = {total_energy}")

# dump to TREXIO file
trexio.to_trexio(mf, 'trexio_diamond.hdf5')
```

### DIFF
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -19,6 +19,7 @@ import pyscf
 from pyscf import lib
 from pyscf import gto
 from pyscf import scf
+from pyscf import pbc
 from pyscf import fci
 
 import trexio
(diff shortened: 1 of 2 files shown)


pyscf/tools/trexio.py
```diff
@@ -200,9 +206,27 @@ def _mol_to_trexio(mol, trexio_file):
...
+            kpt = mol.get_scaled_kpts(kpt)
+            if np.all(np.array(kpt) == 0.0):
+                # gamma point PBC. Real wavefunction
+                trexio.write_pbc_k_point_num(trexio_file, 1)
+                trexio.write_pbc_k_point(trexio_file, [kpt])
+                trexio.write_pbc_k_point_weight(trexio_file, [1.0])
+            else:
+                # general point PBC. Complex wavefunction
+                raise NotImplementedError("Complex WF is not yet implemented.")
+        except ValueError:
```
MatthewRHermes, line 225: A `ValueError` could happen for any number of reasons. I think it's a good idea to check the character of the `ValueError` and make sure it's the specific `ValueError` you are expecting here, before raising an exception with text that may not be accurate.

pyscf/tools/trexio.py
```diff
@@ -200,9 +206,27 @@ def _mol_to_trexio(mol, trexio_file):
...
+            # general points PBC. Complex wavefunctions
+            raise NotImplementedError("Twisted-average is not yet implemented.")
```
MatthewRHermes, line 227: A `ValueError` could happen for any number of reasons. I think it's a good idea to check the character of the `ValueError` and make sure it's the specific `ValueError` you are expecting here, before raising an exception with text that may not be accurate.

MatthewRHermes: Anyway I'm a bit confused because IIUC twist-averaging here would just mean that the k-point is offset from [0,0,0], which would fail the conditional above and lead to the earlier exception. Or does `mol.get_scaled_kpts (mf.kpt)` itself raise the `ValueError`?

sunqm: `mol.get_scaled_kpts (mf.kpt)` can handle any k points. The `ValueError` should never be raised here.

sunqm:
@kousuke-nakano I removed the "twisted-average" error handling code. If this case should be handled in the interface, please add it back in a new PR.

PR #89, opened 2024-12-16
PEP 625

sunqm:
* Fix building issues in aarch64 wheel
* Update sdist name according to PEP 625
* Update PyPI upload configurations

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-NAME = 'pyscf-forge'
+NAME = 'pyscf_forge'
 AUTHOR = 'Pyscf Developer'
 AUTHOR_EMAIL = None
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-NAME = 'pyscf-forge'
+NAME = 'pyscf_forge'
```
MatthewRHermes, line 16: This won't oblige us to rename the repository itself, will it?

sunqm: This change only affects how this project shows on PyPI index page. When using pip, `pip install pyscf-forge` is equivalent to `pip install pyscf_forge`. Pip can do the conversion automatically. I don't it's necessary to change the repository name.

MatthewRHermes (approved):

PR #92, opened 2024-12-20
TREXIO + MCSCF wavefunction

NastaMauger:
Hello,

In this PR, you will find modifications that allow users to register an MCSCF function in the TREXIO format.

I am uncertain if additional data must be stored for TREXIO to function correctly with other software and to return the corresponding energy produced by PySCF before registering the wave function into TREXIO. However, the corresponding HDF5 file generated using the following code:

```
casci = mcscf.casci_symm.CASCI(rhf, norb, nelec)
casci.fcisolver = selected_ci_spin0_symm.SelectedCI(rhf)

casci.kernel()
trexio_file = 'data_casci.hdf5'
pyscf_trexio.to_trexio(casci, trexio_file)

pyscf_trexio.det_to_trexio(casci, norb, nelec, trexio_file)
eri = mol.intor('int2e')
pyscf_trexio.write_eri(eri, trexio_file)
```
seems to register the data correctly.

Please let me know if further modifications or improvements are necessary. I am also uncertain whether the one-electron MO group is essential, as mentione [here.](https://github.com/TREX-CoE/trexio_tools/issues/46#issue-2646327032)

### DIFF
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -21,6 +21,7 @@ from pyscf import lib
 from pyscf import gto
 from pyscf import scf
 from pyscf import pbc
+from pyscf import mcscf
 from pyscf import fci
 
 import trexio
@@ -31,6 +32,8 @@ def to_trexio(obj, filename, backend='h5'):
             _mol_to_trexio(obj, tf)
         elif isinstance(obj, scf.hf.SCF):
             _scf_to_trexio(obj, tf)
+        elif isinstance(obj, mcscf.casci.CASCI) or isinstance(obj, mcscf.CASSCF):
+            _mcscf_to_trexio(obj, tf)
         else:
             raise NotImplementedError(f'Conversion function for {obj.__class__}')
 
@@ -249,7 +252,34 @@ def _cc_to_trexio(cc_obj, trexio_file):
     raise NotImplementedError
 
 def _mcscf_to_trexio(cas_obj, trexio_file):
-    raise NotImplementedError
+    mol = cas_obj.mol
+    _mol_to_trexio(mol, trexio_file)
+    mo_energy_cas = cas_obj.mo_energy
+    mo_cas = cas_obj.mo_coeff
+    num_mo = mo_energy_cas.size
+    spin_cas = np.zeros(mo_energy_cas.size, dtype=int)
+    mo_type_cas = 'CAS'
+    trexio.write_mo_type(trexio_file, mo_type_cas)
+    idx = _order_ao_index(mol)
+    trexio.write_mo_num(trexio_file, num_mo)
+    trexio.write_mo_coefficient(trexio_file, mo_cas[idx].T.ravel())
+    trexio.write_mo_energy(trexio_file, mo_energy_cas)
+    trexio.write_mo_spin(trexio_file, spin_cas)
+
+    ncore = cas_obj.ncore
+    ncas = cas_obj.ncas
+    mo_classes = np.array(["Virtual"] * num_mo, dtype=str)  # Initialize all MOs as Virtual
+    mo_classes[:ncore] = "Core"                    
+    mo_classes[ncore:ncore + ncas] = "Active"     
+    trexio.write_mo_class(trexio_file, list(mo_classes))
+
+    occupation = np.zeros(num_mo) 
+    occupation[:ncore] = 2.0
+    rdm1 = cas_obj.fcisolver.make_rdm1(cas_obj.ci, ncas, cas_obj.nelecas)
+    natural_occ = np.linalg.eigh(rdm1)[0] 
+    occupation[ncore:ncore + ncas] = natural_occ[::-1] 
+    occupation[ncore + ncas:] = 0.0
+    trexio.write_mo_occupation(trexio_file, occupation)
 
 def mol_from_trexio(filename):
     mol = gto.Mole()
@@ -434,10 +464,6 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., c
     with trexio.File(filename, 'u', back_end=_mode(backend)) as tf:
         if trexio.has_determinant(tf):
             trexio.delete_determinant(tf)
-        trexio.write_mo_num(tf, mo_num)
-        trexio.write_electron_up_num(tf, len(a))
-        trexio.write_electron_dn_num(tf, len(b))
-        trexio.write_electron_num(tf, len(a) + len(b))
 
         offset_file = 0
         for i in range(n_chunks):


sunqm (approved):

q-posev:
Thank you @NastaMauger !

I noticed that you do not register the CI determinants as part of your new `mcscf_to_trexio`. Is it intended?

q-posev:
I guess the best test of the validity of the produced multi-configurational TREXIO wavefunction is via an attempt to import and use it in some software that handles such wave functions (e.g. QP2)? @NastaMauger @scemama

pyscf/tools/trexio.py
```diff
@@ -434,10 +464,6 @@ def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., c
     with trexio.File(filename, 'u', back_end=_mode(backend)) as tf:
         if trexio.has_determinant(tf):
             trexio.delete_determinant(tf)
-        trexio.write_mo_num(tf, mo_num)
-        trexio.write_electron_up_num(tf, len(a))
```
q-posev, line 438: Why is this removed? We need the number of up and down electrons to perform the CI determinants "correctness" check before writing them in the file.

NastaMauger:
@q-posev

- _I noticed that you do not register the CI determinants as part of your new mcscf_to_trexio. Is it intended?_
Done

- _I guess the best test of the validity of the produced multi-configurational TREXIO wavefunction is via an attempt to import and use it in some software that handles such wave functions (e.g. QP2)? @NastaMauger @scemama_
There is still an issue due to the normalization factor, which needs to be resolved. The` write-eri` function might still lack consistency between what the original function in pyscf-forge returns and what QP2 (or other software) might expect.

pyscf/tools/trexio.py
NastaMauger: This is now handled by the `_mcscf_to_trexio` function, which calls `_mol_to_trexio`.

q-posev:
@NastaMauger 

For the ERIs - I agree, @kgasperich submitted a [draft PR](https://github.com/pyscf/pyscf-forge/pull/105) which takes care of the AO/MO inconsistency.

NastaMauger:
Hello,

I am following this issue as it's a feature I really want to add. 

After discussing with @q-posev and @scemama, we observed that although we correctly registered all necessary data into TREXIO, there is still a difference between the CASSCF energy from PySCF and the exported TREXIO file. Upon examining the convergence of the energy with respect to the CAS space, it seems there may be a hidden selected CI procedure within the CASSCF routine.

@sunqm  Could you confirm and provide guidance on how to correctly register the CASCi  wavefunction if some selected CI is implicitly performed?

Best

MatthewRHermes:
> Hello,
> 
> I am following this issue as it's a feature I really want to add.
> 
> After discussing with @q-posev and @scemama, we observed that although we correctly registered all necessary data into TREXIO, there is still a difference between the CASSCF energy from PySCF and the exported TREXIO file. Upon examining the convergence of the energy with respect to the CAS space, it seems there may be a hidden selected CI procedure within the CASSCF routine.
> 
> @sunqm Could you please confirm and provide guidance on how to correctly register the CASCI wavefunction if some selected CI is implicitly performed?
> 
> Best

There is no selected CI procedure in CASSCF that I am aware of. There is a procedure to approximate the CI step during the orbital optimization, but that should have no effect on the evaluation of the CASCI energy, which is entirely determined by the MO coefficients, one- and two-electron integrals, and CI vectors. Could you elaborate on what you are seeing?

sunqm:
> Hello,
> 
> I am following this issue as it's a feature I really want to add.
> 
> After discussing with @q-posev and @scemama, we observed that although we correctly registered all necessary data into TREXIO, there is still a difference between the CASSCF energy from PySCF and the exported TREXIO file. Upon examining the convergence of the energy with respect to the CAS space, it seems there may be a hidden selected CI procedure within the CASSCF routine.
> 
> @sunqm Could you please confirm and provide guidance on how to correctly register the CASCI wavefunction if some selected CI is implicitly performed?
> 
> Best

Do reproduce the CASSCF energy for small systems like H2 molecule with (2,2) CAS?

NastaMauger:
I think the differences I got stems for different version I was using. Now it should be ok

scemama (approved):
I am not an expert of PySCF, but on the TREXIO side everything looks OK.

pyscf/tools/trexio.py
```diff
@@ -410,17 +452,19 @@ def get_occsa_and_occsb(mcscf, norb, nelec, ci_threshold=0.):
 
     return occsa_sorted, occsb_sorted, ci_values_sorted, num_determinants
 
-def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., chunk_size=100000):
+def det_to_trexio(mcscf, norb, nelec, trexio_file, ci_threshold=0., chunk_size=100000):
     from trexio_tools.group_tools import determinant as trexio_det
 
     mo_num = norb
+    ncore = mcscf.ncore
     int64_num = int((mo_num - 1) / 64) + 1
```
q-posev, line 460: We can do  `int64_num = trexio.get_int64_num(trexio_file)` as @scemama did in PR #152

NastaMauger: If the user register the determinant is a new file, this will not work

pyscf/tools/trexio.py
```diff
@@ -329,6 +365,12 @@ def write_eri(eri, filename, backend='h5'):
         idx[:,:,2:] = idx_pair[None,:,:]
         idx = idx[np.tril_indices(npair)]
 
+    idx=idx.reshape((num_integrals,4))
```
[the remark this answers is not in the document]
q-posev: How did you test this part? 
I am asking because 
1. I see that you swap the indices to go from one notation into another, but you do not swap the `eri` array accordingly (which remains in the original chemist notation of PySCF). 
2. I see that `read_eri` part is untouched. So when reading the ERI's produced by QP2, they will be read in physicist notation and not in chemist like PySCF may expect it.
I guess one of these points leads to tests failing in the CI. But maybe I missed something?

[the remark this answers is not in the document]
NastaMauger: Regarding your point 2: exactly. That's why I think something should be done in the TREXIO format to ensure that the formats are properly registered and can be used seamlessly across different software.

I never really had the opportunity to finish this, and I know Kevin is working on something similar on the QP2 side.

q-posev (approved):
@NastaMauger thanks! Indeed, the CI was failing before because of the broken ERI import. Now it's fixed and the CI failed only for Python 3.12 for the reasons unknown to me and unrelated to TREXIO, which is good news :-)

So if you ask me, I think this PR can be merged, even though the ERIs are exported in the chemist notation and not in the physicist one at the moment. I believe this issue can be addressed in a separate PR.

What do you think @sunqm @MatthewRHermes @scemama ?

pyscf/tools/trexio.py
scemama: It is possible that in previous versions of trexio you could bypass the need of storing the number of MOs in the trexio file, but it was because some internal consistency checking was lacking in TREXIO. In the current version, you can't write determinants if you don't write the number of MOs in the MO group to ensure that the determinants you store are consistent. So it should be totally equivalent.

I would rather use the library function call than compute it, because it will always return the  value expected by the library. Here it is possible that user makes a mistake, and that norb is not the same as the value stored in the mo group.

scemama:
I think this PR should _not_ be merged until the ERIs are written in phisicist notation, as is specified in the trexio format. Here PySCF is only compatible with itself, are we are back to the problem that trexio was created for. It is dangerous if people start to write wrong trexio files and start to write interfaces from those wrong files!!!

scemama:
> So if you ask me, I think this PR can be merged, even though the ERIs are exported in the chemist notation and not in the physicist one at the moment. I believe this issue can be addressed in a separate PR.

If you want to merge quickly, please re-activate the correct reordering for ERI indices for exporting, and set the importing of ERIs as `NotImplementedError`. And then it can be addressed in a separate PR.

NastaMauger:
Ok @scemama, this is exactly what I was thinking. The correct ERIs are missing, and without them, it's useless to have a QP2 (or any other) test export.

I'm a bit confused about this part: "set the importing of ERIs as NotImplementedError".
Are you asking to create a function called import_eri that simply raises an error?

pyscf/tools/trexio.py
```diff
@@ -329,18 +365,24 @@ def write_eri(eri, filename, backend='h5'):
...
-    '''Read ERIs in AO basis, 8-fold symmetry is assumed'''
     with trexio.File(filename, 'r', back_end=trexio.TREXIO_AUTO) as tf:
         nmo = trexio.read_mo_num(tf)
         nao_pair = nmo * (nmo+1) // 2
         eri_size = nao_pair * (nao_pair+1) // 2
         idx, data, n_read, eof_flag = trexio.read_mo_2e_int_eri(tf, 0, eri_size)
     eri = np.zeros(eri_size)
-    x = idx[:,0]*(idx[:,0]+1)//2 + idx[:,1]
+    x = idx[:,0]*(idx[:,0]+1)//2 + idx[:,2]
     y = idx[:,2]*(idx[:,2]+1)//2 + idx[:,3]
```
q-posev, line 386: @scemama propose to use 
```
    x = idx[:,0]*(idx[:,0]+1)//2 + idx[:,2]
    y = idx[:,1]*(idx[:,1]+1)//2 + idx[:,3]
```
but the second line still has `[:,2]` instead of `[:,1`]. Could it be the reason behind the CI test failure? I am sure it fails because the `read_eri` is broken in this PR, so probably physicist->chemist conversion upon import is not working correctly.

NastaMauger: Oops ...

scemama (approved):

q-posev:
@sunqm We believe that this PR is finally ready to be merged. Would it be possible to get another round of review from one of the `pyscf-forge` maintainers?

kousuke-nakano:
@NastaMauger Hi. I would appreciate it if you could let me know the status of this pull request. I am working on a related topic. #153 Thank you!

NastaMauger:
@kousuke-nakano In fact, this PR is ready to be merged. It is currently blocked for an unknown reason
@sunqm ?

pyscf/tools/trexio.py
```diff
@@ -410,17 +452,18 @@ def get_occsa_and_occsb(mcscf, norb, nelec, ci_threshold=0.):
 
     return occsa_sorted, occsb_sorted, ci_values_sorted, num_determinants
 
-def det_to_trexio(mcscf, norb, nelec, filename, backend='h5', ci_threshold=0., chunk_size=100000):
+def det_to_trexio(mcscf, norb, nelec, trexio_file, ci_threshold=0., chunk_size=100000):
     from trexio_tools.group_tools import determinant as trexio_det
```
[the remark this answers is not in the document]
kousuke-nakano: I want to know your thoughts, @q-posev and @scemama.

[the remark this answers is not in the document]
NastaMauger: Would it be possible to accept this PR as it is and open a new one with your idea? I agree that your suggestion would help reduce dependencies, but since this has already been validated by everyone and the only thing preventing the final merge is unrelated to the current modification, I would really appreciate if it could be merged

To be honest, I’m not sure how much work this would require, especially since this PR has been open for a while simply because I had forgotten about it

kousuke-nakano:
@NastaMauger Yes, I agree with you, and I will create another pull request regarding the dependency issue.

pyscf/tools/trexio.py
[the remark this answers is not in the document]
scemama: > I want to know your thoughts, @q-posev and @scemama.

I agree. We should not import trexio tools.
Also, I recently removed the déterminant module from trexio tools because it was redundant with native functions of trexio.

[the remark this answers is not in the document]
scemama: I don't have my computer with me, so I can't check for sure, but it seems to me that trexio_det is not used, and we can safely remove the import, right?

PR #93, opened 2024-12-20
Translated Meta-GGA on-top functionals and MC23 functional added for the MCPDFT

JangidBhavnesh:
In this PR, I have added.

1. tMGGAs and Hybrid-tMGGAs 
2. MC23 Functional (ChemRxiv. 2024; doi:10.26434/chemrxiv-2024-5hc8g-v2, Accepted in PNAS.)

With this, one can use these functional for MCPDFT energy calculations.

Let me know for any modifications.

@MatthewRHermes 
@matthew-hennefarth


pyscf/mcpdft/otfnal.py
```diff
@@ -253,11 +304,21 @@ def get_ratio (self, Pi, rho_avg):
...
+            Similar for rho_b.
+
+        The translation is done as follows:
 
         rho_t^a = (rho/2) * (1 + zeta)
         rho_t^b = (rho/2) * (1 - zeta)
         rho'_t^a = (rho'/2) * (1 + zeta)
         rho'_t^b = (rho'/2) * (1 - zeta)
+        laplacian_t^a = (laplacian/2) * (1 + zeta)
+        tau_t^a = (tau/2) * (1 + zeta)
```
matthew-hennefarth, line 321: should also just include the `laplacian_t^b` and `tau_t^b` in this doc string.

pyscf/mcpdft/tfnal_derivs.py
```diff
@@ -106,8 +106,7 @@ def eval_ot(otfnal, rho, Pi, dderiv=1, weights=None, _unpack_vot=True):
     assert (rho.shape[0] == 2)
     nderiv = rho.shape[1]
     nderiv_Pi = Pi.shape[0]
-    if nderiv > 4:
-        raise NotImplementedError("Translation of meta-GGA functionals")
+
```
matthew-hennefarth, line 109: Add a new `if nderiv > 6:` even though I don't believe a functional would even exist. Or make it an assert just as a sanity check.

examples/mcpdft/03-metaGGA_functionals.py
```diff
@@ -0,0 +1,41 @@
...
+
+# MC23 = { '0.2952*HF + (1-0.2952)*rep-M06L, 0.2952*HF + (1-0.2952)*rep-M06L'}}
+# To calculate the MC23 energies,
+# First: Calculate the MCPDFT Energies using repM06L
+# Second: Add the CAS Contributation
+
+# State-Specific
+mc = mcpdft.CASCI(mf, 'trepM06L', 6, 8)
+e_pdft = mc.kernel()[0]
+emc23 = 0.2952*mc.e_mcscf + (1-0.2952)*e_pdft
```
matthew-hennefarth, line 30: Ideally the end-user should not have to perform any "hybridization" themselves. What issue are you running into that you cannot set the `ot.hybrid` to `1-0.2952` when you register the new functional?

matthew-hennefarth (commented):
You also need to include unit tests. I would include values from OpenMolcas to ensure that the two codes agree.

examples/mcpdft/03-metaGGA_functionals.py
matthew-hennefarth: are you not able to play around with [these variables](https://github.com/JangidBhavnesh/pyscf-forge/blob/e30dcc418ead44efe84d44149977bb4c067ba9af/pyscf/dft2/libxc.py#L63-L64)?

pyscf/mcpdft/tfnal_derivs.py
```diff
@@ -117,8 +116,21 @@ def eval_ot(otfnal, rho, Pi, dderiv=1, weights=None, _unpack_vot=True):
...
+        rho_deriv = rho_tot[1:4, :]
+    elif nderiv > 4:
+        rho_deriv = rho_tot[1:6, :]
+    else:
+        rho_deriv = None
+
+    if 1 < nderiv_Pi <= 4:
+        Pi_deriv = Pi[1:4, :]
+    elif nderiv > 4:
+        Pi_deriv = Pi[1:6, :]
```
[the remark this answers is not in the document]
matthew-hennefarth: Also, it should be `nderiv_Pi`.

pyscf/mcpdft/tfnal_derivs.py
```diff
@@ -106,8 +106,7 @@ def eval_ot(otfnal, rho, Pi, dderiv=1, weights=None, _unpack_vot=True):
     assert (rho.shape[0] == 2)
     nderiv = rho.shape[1]
     nderiv_Pi = Pi.shape[0]
-    if nderiv > 4:
```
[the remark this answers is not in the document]
matthew-hennefarth: I would also maintain a

```python3
if nderiv_Pi > 4:
    raise NotImplementedError("Translations of unknown functional type")
```

matthew-hennefarth:
Did you try and run a `ftMO6` calculation? Will it crash?

pyscf/mcpdft/tfnal_derivs.py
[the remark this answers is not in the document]
MatthewRHermes: yeah, Pi only has at most 5 rows, and the 5th row  is something else for Pi which is related to the "Colle-Salvetti functional" (which was an early experiment but hasn't been supported for quite some time): 

https://github.com/pyscf/pyscf-forge/blob/6c1acbdcb8a1ac272638221b3e99c38d8fc3492f/pyscf/mcpdft/otpd.py#L75-L79

examples/mcpdft/03-metaGGA_functionals.py
MatthewRHermes: Through some experimentation I figured out the basic recipe for this:

```
from pyscf import dft
from pyscf import dft2
dft.libxc = dft2.libxc
dft.numint.libxc = dft2.libxc
dft.numint.LibXCMixin.libxc = dft2.libxc

# Reparametrized-M06L: rep-M06L
# MC23 = { '0.2952*HF + (1-0.2952)*rep-M06L, 0.2952*HF + (1-0.2952)*rep-M06L'}}
# Here, I am registering the rep-M06L functional only.

# repM06L_C has 27 parameters
MC23_C =  np.array([0.06, 0.0031, 0.00515088, 0.00304966, 2.427648, 3.707473,
                    -7.943377, -2.521466, 2.658691, 2.932276, -8.832841e-01,
                    -1.895247, -2.899644, -5.068570e-01, -2.712838, 9.416102e-02,
                    -3.485860e-03, -5.811240e-04, 6.668814e-04, 0.0, 2.669169e-01,
                    -7.563289e-02, 7.036292e-02, 3.493904e-04, 6.360837e-04, 0.0, 1e-10])

# repM06L_X has 18 parameters
MC23_X =  np.array([3.352197, 6.332929e-01, -9.469553e-01, 2.030835e-01,
                    2.503819, 8.085354e-01, -3.619144, -5.572321e-01,
                    -4.506606, 9.614774e-01, 6.977048, -1.309337, -2.426371,
                    -7.896540e-03, 1.364510e-02, -1.714252e-06, -4.698672e-05, 0.0])

# See: pyscf/pyscf/dft/libxc_funcs.txt
XC_ID_MGGA_C_M06_L = 233
XC_ID_MGGA_X_M06_L = 203

libxc_register_code = 'mc23'.lower ()

dft2.libxc.register_custom_functional_(libxc_register_code, 'M06L',
                                    ext_params={XC_ID_MGGA_X_M06_L: MC23_X,
                                                XC_ID_MGGA_C_M06_L: MC23_C},
                                    hyb=(0.2952,0,0),facs=(0.7048,0.7048))

mc = mcpdft.CASCI (mf, 'tmc23', 6, 8).run ()
```

There is some extra housekeeping you have to do to make this work:
1. switch `mcpdft._libxc` from `dft.libxc` to `dft2.libxc`
2. remove an obvious `NotImplementedError` in `mcpdft.otfnal`
3. make some logic branch or try/except block in `otfnal._hybrid_2c_coeff`

In addition, one probably ought to also
1. Run all the tests, including the usually-skipped `_slow` tests, to make sure none of this broke anything, and
2. Define `'MC23'` in `otfnal.OT_HYB_ALIAS` so that people can use the actual name of the functional in the standard way.l

MatthewRHermes: @Dayou-Zhang the behavior of `dft2.libxc` is a little strange in the code snipped I posted above because if I define `mc23` off of M06 instead of M06-L, placing the hybridization parameter at the appropriate position in the appropriate list, it ends up _adding_ to the hybridization parameter I entered in the `hyb` kwarg and I end up with  overall hyb=0.5904. I decided not to do this because as I said above, hybridization just fundamentally works differently between KS-DFT and MC-PDFT, but I still think this behavior is odd and I don't see in the code how it comes about. Any comments?

JangidBhavnesh: Thanks. 
I will add this.

examples/mcpdft/03-metaGGA_functionals.py
```diff
@@ -0,0 +1,41 @@
...
+print("MC-PDFT state %d Energy at %s OT Functional: %d" % (0, "MC23", emc23))
```
MatthewRHermes, line 31: format the energy as a float (`%f`) not an integer (`%d`)

pyscf/mcpdft/otfnal.py
```diff
@@ -32,6 +32,57 @@
...
+
+    def _register_repM06L():
+        '''
+        This function register the reparametrized
+        M06-L functional (used for MC23 Functional) in the libxc library.
+        '''
+        from pyscf import dft2
+        dft.libxc = dft2.libxc
+        dft.numint.libxc = dft2.libxc
+        dft.numint.LibXCMixin.libxc = dft2.libxc
```
MatthewRHermes, line 55: Could we perhaps do this a little more cleanly? Can we make it so that the entire MC-PDFT module just uses `dft2.libxc` at all times, even when we aren't building these new functionals? Perhaps by intervening in `get_transfnal` and trying to copy some of the logic in `mcdcft.dcfnal.dcfnal.__init__`? This is a highly, highly impure function and it makes me very nervous.

examples/mcpdft/03-metaGGA_functionals.py
Dayou-Zhang: > @Dayou-Zhang the behavior of `dft2.libxc` is a little strange in the code snipped I posted above because if I define `mc23` off of M06 instead of M06-L, placing the hybridization parameter at the appropriate position in the appropriate list, it ends up _adding_ to the hybridization parameter I entered in the `hyb` kwarg and I end up with overall hyb=0.5904. I decided not to do this because hybridization just fundamentally works differently between KS-DFT and MC-PDFT, but I still think this behavior is odd and I don't see in the code how it comes about. Any comments?

Returning hyb=0.5904 is in fact the expected behavior. The `hyb` kwarg overwrites the `hyb` variable returned by `parse_xc` instead of directly setting the total HF exchange. So `hyb[0]` actually represents additional HF exchange that are not part of the LibXC functional components. This can be confirmed by running
```
from pyscf import dft
print(dft.libxc.parse_xc('M06'))
```
which prints
```
((0, 0, 0), ((np.int32(449), 1), (np.int32(235), 1)))
```
We can see `hyb[0]` is zero, not 0.27.

I now realize that the `hyb` kwarg is really counterintuitive. I will try to re-design the API to make it behave more intuitively. Feel free to provide any suggestions if you have any.

pyscf/mcpdft/otfnal.py
Dayou-Zhang: I am also in favor of this suggestion. As a first step, I think the easiest way to avoid impure function is probably override the `libxc` attribute of the `_numint` instance after it is first initialized in the constructors of the on-top functional classes. In the long run, perhaps it would be a good idea to completely remove the `ks` objects from the MC-PDFT module, since all we need is `dft.numint.NumInt` and `dft.grid.Grids`.

pyscf/mcpdft/otfnal.py
JangidBhavnesh: Done

pyscf/mcpdft/tfnal_derivs.py
JangidBhavnesh: I haven't added this line, because we won't be passing rho.shape[1] > 6 to this function.

pyscf/mcpdft/tfnal_derivs.py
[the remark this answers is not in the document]
JangidBhavnesh: I have modified this now.

pyscf/mcpdft/tfnal_derivs.py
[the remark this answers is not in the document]
JangidBhavnesh: I have added this.

examples/mcpdft/03-metaGGA_functionals.py
JangidBhavnesh: Done.

JangidBhavnesh:
> Did you try and run a `ftMO6` calculation? Will it crash?

Yes, it will crash.

examples/mcpdft/03-metaGGA_functionals.py
Dayou-Zhang: I am afraid that supporting strings like `'0.75*mypurefunc+0.25*HF'` is kind of challenging, if at all possible. I would call this "recursively defined functionals" since they are constructed from a previously customized functional.

There are two methods I could think of to make recursively defined functionals possible. In the first method, we do not change the Python API functions such as `eval_xc`. Since all C API functions take an array of `xc_func_type` structs, we need to make a new `xc_func_type` array for the recursively defined functional that copies the `xc_func_type` structs from the previously defined functional component and update coefficients like `hyb` and `fn_facs`. However, LibXC currently does not have an API that performs a deep copy of the functional data structure. We could do a shallow copy of `xc_func_type` with `memcpy`, but this would be problematic if users attempt to set functional parameters on the recursively defined functionals, as it may also change the custom functional that was defined first. Plus this is not memory safe and can cause segfault if the original functional is destructed. In addition, having two functional components with the same integer ID is now possible in recursively defined functionals. The two functional components are likely to be not identical since users could have performed customization on one of them, so we cannot simply merge them by the same integer ID. This would cause the `ext_params` kwarg to fail, since the integer ID is no longer a unique identifier of the functional component.

Alternatively, we can rewrite all the python API functions so that they will call the API functions of the functional components recursively. This would require special handling of the custom functional components separately from the pure LibXC functional components. In this way, we don't need to worry about the two problems in the first method, but it will certainly be much more work. It would be less efficient because of the recursive API call and may also slightly impact the performance of functionals that are not recursively defined.

Because of the complications, I am hesitating whether we should support recursively defined functionals. Since all the functionals are linear combinations, one should probably be able to define any recursively defined functionals in a non-recursive way.

MatthewRHermes: I'm not sure I agree, but we're wandering off topic from Bhavnesh's PR so lets leave it there for now.

pyscf/mcpdft/otfnal.py
```diff
@@ -19,19 +19,110 @@
...
+    try:
+        if xc_code.upper() in REG_OT_FUNCTIONALS:
+            libxc_unregister_code = xc_code.lower()
+            libxc.unregister_custom_functional_(libxc_unregister_code)
+            del REG_OT_FUNCTIONALS[xc_code.upper()]
+
+    except Exception as e:
+        raise RuntimeError(f"Failed to unregister functional '{xc_code}': {e}") from e
+
+def _get_regsitered_ot_functional(xc_code, mol):
```
MatthewRHermes, line 113: typo!

MatthewRHermes:
> > Did you try and run a `ftMO6` calculation? Will it crash?
> 
> Yes, it will crash.

I think it's best if the error message this emits tells the user that fully-translated meta-GGA on-top functionals _are not scientifically defined_; it's not just a matter of derivatives not being propagated far enough. I would raise a `NotImplementedError` in the `ftransfnal` initialization method for any meta-GGA functional.

pyscf/mcpdft/test/test_mgga.py
```diff
@@ -0,0 +1,208 @@
...
+            self.assertAlmostEqual(first, second, expected)
+
+    def test_tmgga(self):
+        e_mcscf = lih_tm06l.e_mcscf
+        epdft = lih_tm06l.e_tot
+
+        sa_e_mcscf = lih_tm06l_sa2.e_mcscf
+        sa_epdft = lih_tm06l_sa2.e_states
+
+        E_CASSCF_EXPECTED = -7.88214917
```
matthew-hennefarth, line 149: Generally I am in favor of including the git commit from which these expected values as a comment. Also it is good to include a comment for if a number came from OpenMolcas or PySCF.

pyscf/mcpdft/tfnal_derivs.py
matthew-hennefarth: it could still be something good to assert as *if* `rho.shape[1] > 6` we have an issue somewhere in the logic of the code (somewhere else).

pyscf/mcpdft/test/test_mgga.py
JangidBhavnesh: I have included at the top of file that these reference values are generated from OpenMolcas, and included the version and other relevant information.

Should I include that information in-front of each reference value?

pyscf/mcpdft/otfnal.py
```diff
@@ -19,19 +19,110 @@
...
+from pyscf.dft2 import libxc
 from pyscf.dft.gen_grid import Grids
 from pyscf.dft.numint import _NumInt, NumInt
 from pyscf.mcpdft import pdft_veff, tfnal_derivs, _libxc, _dms, pdft_feff, pdft_eff
 from pyscf.mcpdft.otpd import get_ontop_pair_density
 from pyscf import __config__
 
+# Points to the libxc of the dft2 module
+dft.numint.libxc = libxc
+dft.numint.LibXCMixin.libxc = libxc
```
MatthewRHermes, line 31: This is still really impure: anything that happens after the `otfnal` module is loaded for the first time will use this modified, "monkey-patched" version of the PySCF core library, even if it's completely unrelated to MC-PDFT. As @Dayou-Zhang said on the previous commit, it's only the otfnal's `_numint` member that actually needs to see this:

```
self._numint.libxc = libxc
```
after lines 298 and 674.

pyscf/mcpdft/tfnal_derivs.py
JangidBhavnesh: Yes. I can add that assertion.

pyscf/mcpdft/otfnal.py
JangidBhavnesh: Corrected.

pyscf/mcpdft/otfnal.py
JangidBhavnesh: Done.

pyscf/mcpdft/tfnal_derivs.py
JangidBhavnesh: I think we don't have to add this assertion as well, because we are [generating](https://github.com/pyscf/pyscf-forge/blob/6c1acbdcb8a1ac272638221b3e99c38d8fc3492f/pyscf/mcpdft/otfnal.py#L80C1-L81C23) the rho, rho', laplacian and tau using 
[dft.numint.eval_rho2 function](https://github.com/pyscf/pyscf/blob/6878b69129e33432bca2249fdfebfd0075f329f2/pyscf/dft/numint.py#L328), which will won't return rho.shape[1]>6.

MatthewRHermes: Yeah, it won't now. But what about later, when other people try to do different things? One should imagine that each function is being written in total isolation and every function that feeds into it is being constantly being edited by a million strangers with a million different objectives.

pyscf/mcpdft/test/test_mgga.py
matthew-hennefarth: I would. The reference values at the top will be forgotten about if others add new tests later in the file which might not come from OpenMolcas anymore. Or if the values start to change over time (for whatever reason). Best to keep the comments as close to the reference values as possible.

pyscf/mcpdft/tfnal_derivs.py
matthew-hennefarth: Yah, it is best to indicate *intent* of the function parameters using `asserts` other `if` statements. This makes it easier to debug later down the line, and so that other users don't try and improperly abuse the functions.

JangidBhavnesh: Sure.

I will add the assertion.

pyscf/mcpdft/test/test_mgga.py
JangidBhavnesh: Done.

pyscf/mcpdft/tfnal_derivs.py
JangidBhavnesh: Done.

MatthewRHermes (changes requested):
One last tweak: see https://github.com/JangidBhavnesh/pyscf-forge/pull/1

JangidBhavnesh:
Done.

MatthewRHermes (approved):
Thanks for all the hard work!

JangidBhavnesh:
Thanks to you, Matthew Hennefarth and Dayou for reviewing it and suggesting necessary changes.

PR #95, opened 2025-01-03
Update noci.py，fix output of dEs and dSplit

baopengbp:


pyscf/msdft/noci.py
```diff
@@ -283,12 +283,13 @@ def multi_states_scf(msks, ground_ks=None):
         log.info('%-2d %18.15g AU %12.6g eV', i+1, e_d, (e_d-e_g)*au2ev)
 
     log.info('Single and triple excitation:')
-    log.info('       E(S)                  E(T)                   dEt               dEs')
+    log.info('             E(S)                  E(T)                   dEt               dEs               dSplit')
```
MatthewRHermes, line 286: Personally, I would use a format string for this. This is more than 100 columns and I'm surprised the lint passed.

sunqm (approved):

PR #97, opened 2025-01-03
Add option for external libs in CMakeLists

jeanwsr:
Recently libxc_itrf2 is added, which links to libxc. But in some cases one may not place the libxc in pyscf/lib/deps. So an option for libxc directory should be added.


MatthewRHermes:
Is this necessary? If LibXC were to be installed externally, wouldn't it naturally already be in LD_LIBRARY_PATH? I don't see this kind of option in the PySCF core module.

jeanwsr:
Yes, it is in the LD_LIBRARY_PATH. But that's not sufficient to make it be recognized by cmake. cmake uses the link_directories() command to determine the path of dynamical lib to be used in "cc -L xxx -l xxx". I'm not sure whether there's any other way to do it, like any cmake options?

In pyscf core program, the CMakeLists add `CMAKE_INSTALL_PREFIX/lib` to link_directories(), which allows users using `-DCMAKE_INSTALL_PREFIX=/path/to/libxc` to specify the libxc path. This is documented in https://pyscf.org/user/install.html#install-without-internet . The option name is a little bit weird because pyscf itself does not use `CMAKE_INSTALL_PREFIX` to install anything. Probably it comes from the `CMAKE_INSTALL_PREFIX` of libcint and libxc in the same installing instruction.

In summary, I think it's necessary to provide the same cmake option as pyscf core do. The name of the option can be discussed: to use XC_DIR as it's simple, or use CMAKE_INSTALL_PREFIX to keep consistency with pyscf, or something else like EXTERNAL_DEPS_DIR.

MatthewRHermes:
I see. I would suggest copying the PySCF core-module solution to this issue if possible, just to minimize confusion.

jeanwsr:
Copied.

MatthewRHermes (approved):
Cool, thanks

PR #98, opened 2025-01-03
Staggered mesh method for PBC exact exchange

stephen-quiton:
Hello! 

In this PR, I'm adding the staggered mesh method for periodic exact exchange ([J. Chem. Theory Comput. 2024, 20, 18, 7958-7968.](https://doi.org/10.1021/acs.jctc.4c00771 "DOI URL")). This is similar in concept to [kmp2_stagger.py](https://github.com/pyscf/pyscf/blob/master/pyscf/pbc/mp/kmp2_stagger.py) but for reducing the finite-size error from the HF exchange term.

Let us know if we need to modify the PR in any way. 

Thank you,
Stephen

### DIFF
--- a/pyscf/pbc/scf/khf_stagger.py
+++ b/pyscf/pbc/scf/khf_stagger.py
@@ -24,7 +24,8 @@ from pyscf.pbc.scf import khf
 from pyscf.pbc.tools.pbc import get_monkhorst_pack_size
 
 
-def exchange_energy_stagger(kmf, stagger_type=2, df_type=None, kshift_rel=0.5, verbose=logger.NOTE, with_vk=False, nks=None):
+def exchange_energy_stagger(kmf, stagger_type=2, df_type=None, kshift_rel=0.5, verbose=logger.NOTE,
+                            with_vk=False, nks=None):
     """ Computes the exchange energy using the staggered mesh method based on
         J. Chem. Theory Comput. 2024, 20, 18, 7958-7968.
 
@@ -221,8 +222,9 @@ class KHF_stagger(khf.KSCF):
             ej = 1. / Nk * np.einsum('kij,kji', Jo, dm_kpts) * 0.5
             self.ej = ej.real
         if k:
-            results = exchange_energy_stagger(self.cell, self.kpts, stagger_type=self.stagger_type, df_type=self.df_type,
-                                              dm_kpts=self.dm_kpts, mo_coeff_kpts=self.mo_coeff_kpts, kshift_rel=self.kshift_rel,
+            results = exchange_energy_stagger(self.cell, self.kpts, stagger_type=self.stagger_type,
+                                              df_type=self.df_type, dm_kpts=self.dm_kpts,
+                                              mo_coeff_kpts=self.mo_coeff_kpts, kshift_rel=self.kshift_rel,
                                               with_vk=self.with_vk)
             self.ek = results["E_stagger_M"]
         if xc:
@@ -319,7 +321,8 @@ class KHF_stagger(khf.KSCF):
         prev = 0
         converged_madelung = 0
         while True and kmf.cell.dimension !=1:
-            madelung = KHF_stagger.modified_madelung(cell_input=ecell, kshift_abs=kshift_abs, ew_eta=ew_eta, ew_cut=ew_cut)
+            madelung = KHF_stagger.modified_madelung(cell_input=ecell, kshift_abs=kshift_abs, ew_eta=ew_eta,
+                                                     ew_cut=ew_cut)
             log.debug1("Iteration number " + str(count_iter))
             log.debug1("Madelung:" + str(madelung))
             log.debug1("Eta:" + str(ew_eta))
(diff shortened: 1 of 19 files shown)


pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+        ecell = copy.copy(mf.cell)
+        ecell._atm = np.array([[1, mf.cell._env.size, 0, 0, 0, 0]])
+        ecell._env = np.append(mf.cell._env, [0., 0., 0.])
+        ecell.unit = 'B'
+
+        # Expand the unit cell to the supercell based on the Monkhorst-Pack grid size
+        ecell.a = np.einsum('xi,x->xi', mf.cell.lattice_vectors(), nks)
+        ecell.mesh = np.asarray(mf.cell.mesh) * nks
+
+        print('build_probe_cell: nks ',nks)
```
MatthewRHermes, line 76: Could you use the logger instead of "print"?

pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+def kernel(kmf, type="Non-SCF", df_type=None, kshift_rel=0.5, verbose=logger.NOTE, with_vk=False, nks=None):
+    """ Computes the exchange energy using the staggered mesh method based on
+        J. Chem. Theory Comput. 2024, 20, 18, 7958-7968.
+
+    Args:
+        kmf (khf.KSCF): KSCF object (KRKS or KRHF supported). For Non-SCF and
+            Split-SCF, the SCF should be converged.
+        type (str, optional): "Regular", "Non-SCF", or "Split-SCF". See
```
MatthewRHermes, line 34: Not quite: this actually requires an integer, since you don't call `get_stagger_type_id` in this function. You only call it in the class kernel method.

pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+            self.exc = 0.0
+            hyb = 1.0
+        else:
+            logger.error("KHF Stagger: Invalid SCF type", self.mf.__class__)
+            raise ValueError("Invalid SCF type")
+
+        self.compute_energy_components(hcore=True,nuc=True,j=True,k=False,xc=xc)
+        self.e_tot = hyb * self.ek + self.ehcore + self.enuc + self.ej + self.exc
+        self.dump_flags()
+        return self.e_tot
```
MatthewRHermes, line 376: Ideally, we preserve the calling arguments and returns of the parent function. What if I have a guess density matrix I want to use? Etc.

pyscf/pbc/scf/khf_stagger.py
stephen-quiton: These were debugging print statements that should have been deleted. They're fixed now; thanks!

pyscf/pbc/scf/khf_stagger.py
stephen-quiton: Thank you for the catch! [It should be corrected now.](https://github.com/pyscf/pyscf-forge/blob/82882a9f3e495b31c00ee2373b1baa75f1e73cc0/pyscf/pbc/scf/khf_stagger.py#L35)

pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+                Madelung-like correction.
+            E_stagger (float): Staggered mesh exchange energy.
+            vk (np.array, optional): Exchange matrix if with_vk=True.
+    """
+
+    icell = kmf.cell
+    log = logger.Logger(icell.stdout, verbose)
+    ikpts = kmf.kpts
+
+    def build_probe_cell(mf, nks=None):
```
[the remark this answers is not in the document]
stephen-quiton: Good question, I think it's because when I was first implementing these functions, they had variables that were only defined in that `kernel` function (now called `exchange_energy_stagger`), but [I've now moved them inside the `KHF_stagger` class.](https://github.com/pyscf/pyscf-forge/blob/82882a9f3e495b31c00ee2373b1baa75f1e73cc0/pyscf/pbc/scf/khf_stagger.py#L292-L369)

pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+
+        # Diagonalize to get densities at shifted mesh
+        mo_energy_shift, mo_coeff_shift = kmf.eig(F_shift, s1e)
+        mo_occ_shift = kmf.get_occ(mo_energy_kpts=mo_energy_shift, mo_coeff_kpts=mo_coeff_shift)
+        dm_shifted = kmf.make_rdm1(mo_coeff_kpts=mo_coeff_shift, mo_occ_kpts=mo_occ_shift)
+
+        # Compute the Staggered mesh energy
+        Nk = np.prod(nks)
+        E_stagger = -1./Nk * np.einsum('kij,kji', dm_shifted, Kmat) * 0.5
+        E_stagger /= 2.
```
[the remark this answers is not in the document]
stephen-quiton: [I've refactored it so it's more compact and readable.](https://github.com/pyscf/pyscf-forge/blob/82882a9f3e495b31c00ee2373b1baa75f1e73cc0/pyscf/pbc/scf/khf_stagger.py#L79-L122) Thanks for pointing this out!

pyscf/pbc/scf/khf_stagger.py
```diff
@@ -0,0 +1,420 @@
...
+from pyscf.lib import logger
+from pyscf.pbc import df
+from pyscf.pbc import scf as pbcscf
+from pyscf.pbc.dft import numint as pbcnumint
+from pyscf.pbc.dft import rks
+from pyscf.pbc.scf import khf
+from pyscf.pbc.tools.pbc import get_monkhorst_pack_size
+
+
+def kernel(kmf, type="Non-SCF", df_type=None, kshift_rel=0.5, verbose=logger.NOTE, with_vk=False, nks=None):
```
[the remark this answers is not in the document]
stephen-quiton: Makes sense, [changed to `exchange_energy_stagger`](https://github.com/pyscf/pyscf-forge/blob/82882a9f3e495b31c00ee2373b1baa75f1e73cc0/pyscf/pbc/scf/khf_stagger.py#L27)

pyscf/pbc/scf/khf_stagger.py
stephen-quiton: Great point, [I've made it so that the user can pass dm0 and other args to kernel,](https://github.com/pyscf/pyscf-forge/blob/82882a9f3e495b31c00ee2373b1baa75f1e73cc0/pyscf/pbc/scf/khf_stagger.py#L340) which will be used for the regular version of staggered mesh.

stephen-quiton:
Thank you for the feedback @MatthewRHermes! I believe I've addressed your comments, but please let me know if any other changes need to be made or if I missed anything.

sunqm:
The error in CI is very strange. Can you merge the head?

stephen-quiton:
> The error in CI is very strange. Can you merge the head?
Done!

PR #99, opened 2025-01-08
Sanity Check of the mol object in MCPDFT

JangidBhavnesh:
In the current implementation, using a CAS object calculated at the $$\Gamma$$-point (PBC) for MCPDFT does not raise an error. However, in the case of periodic boundary conditions, one needs to use specific quadrature grids, which have not been implemented yet. Therefore, in this PR, I have added a function that checks the `mol` object during the initialization of the MCPDFT object and raises an error if a `cell` object is found.

@MatthewRHermes

### DIFF
--- a/pyscf/mcpdft/__init__.py
+++ b/pyscf/mcpdft/__init__.py
@@ -21,7 +21,7 @@ from pyscf import mcscf, gto
 from pyscf.lib import logger
 from pyscf.mcscf import mc1step, casci
 
-def _sanity_check__of_mol(mc_or_mf_or_mol):
+def _sanity_check_of_mol(mc_or_mf_or_mol):
     '''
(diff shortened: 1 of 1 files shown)


pyscf/mcpdft/__init__.py
```diff
@@ -21,11 +21,29 @@
 from pyscf.lib import logger
 from pyscf.mcscf import mc1step, casci
 
+def _sanity_check__of_mol(mc_or_mf_or_mol):
```
matthew-hennefarth, line 24: small spelling `_sanity_check__of_mol` -> `_sanity_check_of_mol`

JangidBhavnesh: Fixed. 

Thanks

MatthewRHermes (approved):
Thank you @JangidBhavnesh

Issue #103, opened 2025-01-17
trexio possible issue with ERI basis


kgasperich:
I've noticed something that seems a bit unclear (and hopefully this post might provide clarity for others).

The [`trexio.write_eri`](https://github.com/pyscf/pyscf-forge/blob/0c2586886d890bae40eb4833151f9527c42f90bd/pyscf/tools/trexio.py#L310-L333) function always writes the input ERIs as if they are in the MO basis (i.e. it writes them under `mo_2e_int` in the trexio h5 file), but in some of the tests, the AO ERIs are passed to this function ([1](https://github.com/pyscf/pyscf-forge/blob/0c2586886d890bae40eb4833151f9527c42f90bd/pyscf/tools/test/test_trexio.py#L76), [2](https://github.com/pyscf/pyscf-forge/blob/0c2586886d890bae40eb4833151f9527c42f90bd/pyscf/tools/test/test_trexio.py#L87)), so when used in this way, the trexio h5 output will contain the AO ERIs labeled incorrectly as MO ERIs.

It might be clearer to rename the function to `write_mo_eri` or maybe to provide another argument to select AOs or MOs?
It might also make sense to make a function that takes the scf object, and that will contain everything needed to get the ERIs in either basis.
My usual workflow is to get the AO ERIs with `mol.intor('int2e')` and then `ao2mo.kernel(eri_ao, mf.mo_coeff)` to get the MO ERIs if needed, but I wouldn't be surprised if there were maybe cleaner ways to do this that I'm not aware of (e.g. I didn't know about `scf._eri`)

sunqm:
There is no essential difference between AO ERIs and MO ERIs. AO ERIs can be viewed as MO ERIs with the MO coefficients set to the identity matrix. In the tests, AO ERIs are passed to this function to verify the handling of indices. It does not indicate special meaning for the input ERIs. Does TrexIO provide different entries or labels for AO and MO ERIs? If not, I don't think it's necessary to distinguish the two when dumping ERIs. They can be easily converted to each other before calling this function.

kgasperich:
Yes, TREXIO does provide different labels for [AO](https://trex-coe.github.io/trexio/trex.html#ao_two_e) and [MO](https://trex-coe.github.io/trexio/trex.html#org6f12366) ERIs (without this distinction, I agree that there is no real difference between the two other than the basis in which they're represented, and it would be up to the user to pay attention and convert as needed).

[...]

PR #105, opened 2025-01-20
[WIP] clearer trexio eri interface

kgasperich:
This is just one example of a way to address the issues brought up in Issue #103 

The changes here are just the first implementation I thought of that adds some clarity/safety, so I'm not strongly attached to the way I've done it here, and someone more familiar with pyscf might know of a cleaner way to handle it, especially regarding the `ao2mo` part.  I'm not familiar with all of the possible options for input/output symmetries, so it might make sense to pass some more args through to that function to allow for various possibilities.

I also haven't changed/added any tests here, but I could do that if there is any interest in merging the PR.

### DIFF
--- a/pyscf/tools/trexio.py
+++ b/pyscf/tools/trexio.py
@@ -22,6 +22,7 @@ from pyscf import gto
 from pyscf import scf
 from pyscf import pbc
 from pyscf import fci
+from pyscf import ao2mo
 
 import trexio
 
@@ -307,7 +308,8 @@ def scf_from_trexio(filename):
     mf.mo_occ = mo_occ
     return mf
 
-def write_eri(eri, filename, backend='h5'):
+def write_eri(eri, filename, backend='h5', basis='mo'):
+    assert basis.upper() in ['MO','AO']
     num_integrals = eri.size
     if eri.ndim == 4:
         n = eri.shape[0]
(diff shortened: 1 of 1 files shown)


NastaMauger:
Hello @kgasperich 

I think to work effectively, we might want to add a chemist/physicist notation feature, either directly in TREXIO or within the PySCF-Tools. This would help standardize how we handle the ERIs and ensure clarity when dealing with different notation.

In my [PR](https://github.com/pyscf/pyscf-forge/pull/92). after the `write_eri` function, I added the following:

    idx=idx.reshape((num_integrals,4))
    for i in range(num_integrals):
        idx[i,1],idx[i,2]=idx[i,2],idx[i,1] 

    idx=idx.flatten()

    with trexio.File(filename, 'w', back_end=_mode(backend)) as tf: 
        trexio.write_mo_2e_int_eri(tf, 0, num_integrals, idx, eri.ravel())


I believe this change should be incorporated. Let me know your thoughts.

Best

kousuke-nakano:
Dear @kgasperich, thank you for the effort! We are now finishing MCSCF implementation #92. Once it is merged, could we work on this topic together? I agree that we should also add `ao_2e`. Do you think it is worth supporting all the combinations of integrals? i.e., `mo_2e`, `mo_1e`, `ao_2e`, and `ao_1e`?

PR #106, opened 2025-01-22
Fix pypi release pipeline

sunqm:


MatthewRHermes:
I'm not familiar enough yet with pypi release pipelines to evaluate this

PR #108, opened 2025-02-17
Fix PDFT effective potential numerical instability

matthew-hennefarth:
Translated GGA functionals `tN12` and `tGAM` (and other functionals relate to a b97 correlation function) had a numerical instability when calculation the 1- and 2- particle effective potential terms for systems with no beta electrons. This was related to the `idx` in `get_ratio`. This PR fixes the issue and adds tests to the `get_pdft_veff` and `get_pdft_feff` to prevent regression.


matthew-hennefarth:
There seems to be more numerical instability in the `feff` term. On my machine, this passes with no problem, but it seems to be failing for python 3.10.

MatthewRHermes (approved):

PR #111, opened 2025-02-24
Debug for sftda.uhf_df.get_ab_sf() function when the system is large.

Haskiy:
When system is large, the function block_loop will divides grids into several parts of to calculate. Before, we didn't note this. At this time, we correct the sftda.uhf_df.get_ab_sf() function to meet this setting. And there may be a disadvantage for mc approach that functional derivatives at all grids are required to save in the calculation.


Haskiy:
I think this failure originates from the tests of mcpdft module. Can I solve it?

matthew-hennefarth:
> I think this failure originates from the tests of mcpdft module. Can I solve it?

It might be related to #112 and PR #113.

Issue #112, opened 2025-02-24
PR #108 broke total energies of H2 triplet


MatthewRHermes:
@matthew-hennefarth 

I'm sorry, I really should have checked this before accepting, but it looks like we don't actually have any unit test anywhere that evaluates the **total energy** of a molecule like H2 triplet that has no beta electrons.

Before PR #108, the output of this:

```
from pyscf import gto, scf
from pyscf import mcpdft

mol = gto.M (atom='H 0 0 0; H 1 0 0', basis='sto-3g', spin=2)
mf = scf.RHF (mol).run ()
mc = mcpdft.CASSCF (mf, 'tPBE', 2, 2).run ()
mc = mcpdft.CASSCF (mf, 'ftPBE', 2, 2).run ()
```
is this:
```
converged SCF energy = -0.745871793035567
CASSCF energy = -0.745871793035567
CASCI E = -0.745871793035567  E(CI) = -1.27504900395557  S^2 = 2.0000000
MC-PDFT state 0 E = -0.7470290907770314, Eot(tPBE) = -0.8376845044629804
CASSCF energy = -0.745871793035567
CASCI E = -0.745871793035567  E(CI) = -1.27504900395557  S^2 = 2.0000000
MC-PDFT state 0 E = -0.7470290907770314, Eot(ftPBE) = -0.8376845044629804
```
which agrees with OpenMolcas 24.10, which using ultrafine grid has it as -0.74702903. However, after PR #108, the output is:
```
converged SCF energy = -0.745871793035567
CASSCF energy = -0.745871793035567
CASCI E = -0.745871793035567  E(CI) = -1.27504900395557  S^2 = 2.0000000
MC-PDFT state 0 E = -0.626825259982524, Eot(tPBE) = -0.717480673668473
CASSCF energy = -0.745871793035567
CASCI E = -0.745871793035567  E(CI) = -1.27504900395557  S^2 = 2.0000000
MC-PDFT state 0 E = -0.626825259982524, Eot(ftPBE) = -0.717480673668473
``` 
The fact that this breaks makes sense upon reflection, because the index down here which omits columns with Pi == 0:
https://github.com/pyscf/pyscf-forge/blob/e9f8187b24fa3f251e4975f42fe002d8bc9dc774/pyscf/mcpdft/otfnal.py#L334-L341
[...]

PR #113, opened 2025-02-24
Repair #108 and add all-spin-up unittest (#112)

MatthewRHermes:
PR #108 inadvertently caused all MC-PDFT calculations of molecules with all spin-up electrons to produce incorrect energies, which was not caught at the time because it turns out we had no actual unit tests of the **values** of the energies in that specific edge case.


MatthewRHermes:
Tests originally added by #108 now re-enabled. @matthew-hennefarth see commit message above

PR #117, opened 2025-02-26
Translated meta-GGA on-top potentials

matthew-hennefarth:
This PR:
- Updates MC-PDFT docs to include new meta-GGA functionality with references (and updates some references).
- Fixes `rho` generators to no longer compute the laplacian of rho, and only uses $\tau$ for meta-GGA functionals.
- Implements the Jacobian for translated meta-GGA functionals to compute $\frac{\partial E^\mathrm{ot}}{\partial \gamma^p_q}$ and $\frac{\partial E^\mathrm{ot}}{\partial \gamma^{pr}_{qs}}$ (the one- and two-particle on-top potentials). 
  - Also appends `tM06L` as a test to `test_pdft_veff`
- Adds test for L-PDFT with MC23 functional.
- Fixes `test_otfnal` to also check meta-GGA functionals


pyscf/mcpdft/tfnal_derivs.py
```diff
@@ -86,8 +86,9 @@ def eval_ot(otfnal, rho, Pi, dderiv=1, weights=None, _unpack_vot=True):
             first functional derivative of Eot wrt (density, pair
             density) and their derivatives. If _unpack_vot = True, shape
             and format is ([a, ngrids], [b, ngrids]) : (vrho, vPi);
-            otherwise, [c, ngrids] : [rho,Pi,|rho'|^2,rho'.Pi',|Pi'|^2]
+            otherwise, [c, ngrids] : [rho,Pi,|rho'|^2,tau,rho'.Pi',|Pi'|^2]
             ftGGA: a=4, b=4, c=5
+            tmGGA: a=5, b=1, c=4 (drop Pi')
             tGGA: a=4, b=1, c=3 (drop Pi')
             *tLDA: a=1, b=1, c=2 (drop rho')
```
MatthewRHermes, line 93: Can we clarify "drop tau" in ftGGA, tGGA, and *tLDA?

pyscf/mcpdft/tfnal_derivs.py
```diff
@@ -455,6 +481,34 @@ def _tGGA_jT_op(x, rho, Pi, R, zeta):
...
+    jTx = np.zeros((4, ngrid), dtype=x[0].dtype)
+    if R.ndim > 1:
+        R = R[0]
+
+    # ab -> cs coordinate transformation
+    xc = (x[5] + x[6]) / 2.0
+    xm = (x[5] - x[6]) / 2.0
+
+    # easy part
+    jTx[3] = xc + zeta[0] * xm
```
MatthewRHermes, line 497: Is this vulnerable to the problem that #113 solved? We haven't seen it happen yet but is it possible that some functional might return nonsense on x[6] when there are no beta electrons?

MatthewRHermes: wait nevermind, thats what 121-129 are for

matthew-hennefarth: Yes, if I don't have the 121-129 lines to prevent 0 beta electrons, I will get non-sensical $\frac{\partial \epsilon^\mathrm{xc}}{\tau_\downarrow}$. Specifically, this is an issue with the $\frac{\partial}{\partial \Pi}$ term.

pyscf/mcpdft/tfnal_derivs.py
matthew-hennefarth: Fixed.

MatthewRHermes (approved):

pyscf/mcpdft/tfnal_derivs.py
MatthewRHermes: One could make the argument that we ought to have "defense in depth" for this sort of thing by both making sure we never have identically zero translated beta electrons and by indexing explicitly for zeta being identically one in this function, in case a situation ever arises where 1e-15 is not far enough away from zero to control instability. But IDK if it's that important.

matthew-hennefarth: Its actually not this step that causes an issue. I have tried indexing off of `zeta==1` and setting the default to `x[5]` but this doesn't solve the problem. The problem, for H2 triplet, tM06L functional, is actually on the `jTx[1]` term, or the derivative with respect to $\Pi$. If I write it out exactly, we have that 

$$
  \frac{2\tau}{\rho^2} \frac{\partial\zeta}{\partial R} \left( \frac{\partial \epsilon^\mathrm{xc}}{\tau_\uparrow} - \frac{\partial \epsilon^\mathrm{xc}}{\tau_\downarrow}\right)
$$

At $R = 0$, $\zeta = 1$ and $\frac{\partial\zeta}{\partial R} = -\frac{1}{2}$. So if $\frac{\partial \epsilon^\mathrm{xc}}{\tau_\downarrow}$ gives non-sensible values because of whatever issue is present in libxc, there is no multiplier out front to 'kill' this term off.

MatthewRHermes:
Thanks for the hard work!

Issue #119, opened 2025-03-03
Gradient task of CMS-PDFT failed after CASSCF gradient finished.


Sang-YQ:
Hi, developers!

The gradient of aniline at the level of CMS-PDFT(14,14) failed.

Input file:
```
from pyscf import mcpdft
mc = mcpdft.CASSCF(mf, 'ftPBE', 14, 14)
mc = mc.multi_state([1/3, 1/3, 1/3], 'cms')
mc.chkfile = 'cms-T3.chk'
mc.fix_spin_(ss=2)
mc.natorb = True
mc = df.density_fit(mc)
mc.kernel()

g = mc.nuc_grad_method()
for i in range(3):
    g.kernel(state=i)
```

Error information:
```
Traceback (most recent call last):
  File "/shared_storage/home/sangyq/aniline/cms-T3.py", line 77, in <module>
    g.kernel(state=i)
  File "/shared_apps/anaconda3/envs/env_mokit/lib/python3.11/site-packages/pyscf/df/grad/mspdft.py", line 35, in kernel
    return mspdft_grad.Gradients.kernel (self, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/shared_apps/anaconda3/envs/env_mokit/lib/python3.11/site-packages/pyscf/grad/mspdft.py", line 271, in kernel
    return mcpdft_grad.Gradients.kernel (self, state=state, mo=mo, ci=ci,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/shared_apps/anaconda3/envs/env_mokit/lib/python3.11/site-packages/pyscf/grad/mcpdft.py", line 517, in kernel
    return super().kernel (**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/shared_apps/anaconda3/envs/env_mokit/lib/python3.11/site-packages/pyscf/grad/sacasscf.py", line 582, in kernel
    return lagrange.Gradients.kernel (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/shared_apps/anaconda3/envs/env_mokit/lib/python3.11/site-packages/pyscf/grad/lagrange.py", line 159, in kernel
    ham_response = self.get_ham_response (**kwargs)
[...]

PR #121, opened 2025-03-04
Handling different types of mean-field bases in MC-PDFT calculations

MatthewRHermes:
Will solve #119 (see thread for workaround in the meantime).

TODO:

- [x] add test coverage for high-spin ROHF-based MC-PDFT calculations
- [x] determine how to handle UHF bases. It's a prerequisite if we ever hope to work with 2-component or 4-component generalizations of MC-PDFT ([such as](https://pubs.acs.org/doi/abs/10.1021/acs.jctc.2c00062)). Is there any version of MC-PDFT for which this currently "just works"?


MatthewRHermes:
OK, UHF-based MC-PDFT behaves exactly like PySCF core-module mcscf, where it just switches over to RHF silently unless you type "UCASSCF" or "UCASCI", and those aren't defined yet for the mcpdft module. That's fine.

sunqm (approved):

PR #131, opened 2025-04-05
change the davidson solver for spin flip TDDFT

Haskiy:
Change the davidson solver for spin flip TDDFT, where lib.davidson_nosym1() is used. 
Unlike the case of spin-conserving TDDFT, the matrix in the Casida equation for spin-flip TDDFT (not TDA) is non-Hermitian, so the davidson_nosym1() solver should be used. The following updates have been made:
1. Changed davidson1() to davidson_nosym1() in sftda.uks_sf.kernel();
2. Added unit tests and usage examples for spin-flip TDDFT: (1) sftda.test.test_uks_sf_tddft.py；(2) example.grad.04-spin_flip_tddft_grad.py.


MatthewRHermes (commented):
Can we get a brief comment about why this is being done?

Haskiy:
> Can we get a brief comment about why this is being done?
Thank you for the reminder. I have revised the content of the first comment to clarify the reason for this code update.

MatthewRHermes (approved):

PR #135, opened 2025-05-03
Add a feature tranfer template

sunqm:
@gkc1000 @ssh2 @tberkel @sanshar any suggestions for the template?


gkc1000:
Can we request that authors provide their authorship info (email address/organization if applicable) at this stage.

PR #140, opened 2025-05-05
Import CSF from correct target

MatthewRHermes:
CSF has been moved to pyscf-forge. No need to import it from mrh!

### DIFF
--- a/pyscf/mcpdft/lpdft.py
+++ b/pyscf/mcpdft/lpdft.py
@@ -731,8 +731,8 @@ def linear_multi_state_mix(mc, fcisolvers, weights=(0.5, 0.5), **kwargs):
 
 if __name__ == "__main__":
     from pyscf import gto, scf
-    from mrh.my_pyscf import mcpdft
-    from mrh.my_pyscf.fci import csf_solver
+    from pyscf import mcpdft
+    from pyscf.csf_fci import csf_solver
 
     mol = gto.M(
         atom="""H 0 0 0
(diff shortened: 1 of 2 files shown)


pyscf/mcpdft/lpdft.py
```diff
@@ -731,8 +731,8 @@ class LPDFT(_LPDFTMix, mcbase_class):
 
 if __name__ == "__main__":
     from pyscf import gto, scf
-    from mrh.my_pyscf import mcpdft
-    from mrh.my_pyscf.fci import csf_solver
+    from pyscf import mcpdft
```
matthew-hennefarth, line 734: Honestly, this block of code (the `if __name__ == "__main__":`) could just be removed. It is not exactly needed, and `lpdft.py` should not be called directly.

matthew-hennefarth: They can decay quickly and have code that doesn't work with the current code base (there is no CI/CD for these strips of code). Current PySCF documentation takes from either doc strings of functions or examples to provide additional information to users in the published documentation online.

MatthewRHermes: I mean, overall, the question of whether to keeps these things at all or not seems like a separate and bigger issue than the topic of this PR.

sunqm: Some style guidelines recommend removing the `if __name__ == '__main__'` block. Newly developed pyscf modules mostly do not have this block. I think it's fine to keep it as long as it's simple and easy to understand. Feel free to leave it in if you prefer.

Issue #128, opened 2025-03-14
Better get_energy_decomposition for MCPDFT


jeanwsr:
The `get_energy_decomposition` function shares part of its decomposition functionality with other functions, like `energy_mcwfn`. However, the energy components outputed by them is not unified. For example, `energy_mcwfn` outputs `E_x` of mcwfn at `verbose=DEBUG` level, but `get_energy_decomposition` never computes that. I believe it's better to unify the behavior of them, or simply reuse some codes.

Moreover, it may be better to return a dict (like the `mf.scf_summary` in pyscf's HF) than a fixed number of variables in `get_energy_decomposition`, since with dict we will not break the backward compatibility when adding a new component in the future.

I can submit a PR if you are supportive of these enhancements.

matthew-hennefarth:
> For example, `energy_mcwfn` outputs `E_x` of mcwfn at `verbose=DEBUG` level, but `get_energy_decomposition` never computes that. 

Since MC-PDFT is supposed to work with a multiconfiguration wave function, exchange energy is not well defined, hence why it does not compute that. That is not to say the "exchange integrals" are ill-defined, but they don't represent the exchange energy of the system.

> I believe it's better to unify the behavior of them, or simply reuse some codes.

If there are opportunities to compute parts of the wave function energy using code in the main repo, sure I don't see why not. Should help with maintainability. 

> Moreover, it may be better to return a dict (like the `mf.scf_summary` in pyscf's HF) than a fixed number of variables in `get_energy_decomposition`, since with dict we will not break the backward compatibility when adding a new component in the future.

I can see this being a useful addition/modification. Of course, then we need to make sure we use a consistent naming scheme for the energy variables.

jeanwsr:
> Since MC-PDFT is supposed to work with a multiconfiguration wave function, exchange energy is not well defined, hence why it does not compute that. That is not to say the "exchange integrals" are ill-defined, but they don't represent the exchange energy of the system.

Yes, it is not well defined. But since this component has already been computed by `energy_mcwfn` and counted in the total energy as `(hyb_x * E_x)`, I believe it can be a reasonable component to be included in `get_energy_decomposition` (we can point out it's not a well-defined E_x in doc). Also it's useful to have this returned when users want to tuning the `hyb_x` and `hyb_c` parameters.

matthew-hennefarth:
> Also it's useful to have this returned when users want to tuning the `hyb_x` and `hyb_c` parameters after the mcpdft job.

[...]

PR #141, opened 2025-05-05
Remove physically incorrect debug information in MC-PDFT

matthew-hennefarth:
See #128. Removes incorrect debug information that can lead to confusion.

Also warns users and points to the discussion if exchange and correlation are not scaled at the same amount. Perhaps this warning should eventually moved into `otfnal`.

### DIFF
--- a/pyscf/mcpdft/mcpdft.py
+++ b/pyscf/mcpdft/mcpdft.py
@@ -147,12 +147,13 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
     if log.verbose >= logger.DEBUG or abs(hyb_x) > 1e-10:
         vj, vk = mc._scf.get_jk(dm=dm1s)
         vj = vj[0] + vj[1]
+
     else:
         vj = mc._scf.get_j(dm=dm1)
+
     Te_Vne = np.tensordot(h, dm1)
     # (vj_a + vj_b) * (dm_a + dm_b)
     E_j = np.tensordot(vj, dm1) / 2
-    # (vk_a * dm_a) + (vk_b * dm_b) Mind the difference!
     log.debug('CAS energy decomposition:')
     log.debug('Vnn = %s', Vnn)
     log.debug('Te + Vne = %s', Te_Vne)
@@ -160,11 +161,12 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
 
     if abs(hyb_x - hyb_c) > 1e-10:
         log.warn("exchange and correlation hybridization differ")
-        log.warn("Your energy surfaces may have discontinuities, see https://github.com/pyscf/pyscf-forge/issues/128")
+        log.warn("may lead to unphysical results, see https://github.com/pyscf/pyscf-forge/issues/128")
 
     # Note: this is not the true exchange energy, but just the HF-like exchange
     E_x = 0.0
     if log.verbose >= logger.DEBUG or abs(hyb_x) > 1e-10:
+        # (vk_a * dm_a) + (vk_b * dm_b)
         E_x = -(np.tensordot(vk[0], dm1s[0]) + np.tensordot(vk[1], dm1s[1]))
         E_x /= 2.0
         log.debug("E_x = %s", E_x)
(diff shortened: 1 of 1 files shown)


pyscf/mcpdft/mcpdft.py
```diff
@@ -153,26 +153,30 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
...
-        E_x /= 2.0
-    else:
-        E_x = 0
     log.debug('CAS energy decomposition:')
     log.debug('Vnn = %s', Vnn)
     log.debug('Te + Vne = %s', Te_Vne)
     log.debug('E_j = %s', E_j)
-    log.debug('E_x = %s', E_x)
+
+    if abs(hyb_x-hyb_c) < 1e-10:
```
jeanwsr, line 161:
```suggestion
    if abs(hyb_x-hyb_c) > 1e-10:
```
typo?

pyscf/mcpdft/mcpdft.py
```diff
@@ -153,26 +153,30 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
...
     E_c = 0
     if log.verbose >= logger.DEBUG or abs(hyb_c) > 1e-10:
-        # g_pqrs * l_pqrs / 2
-        # if log.verbose >= logger.DEBUG:
         aeri = ao2mo.restore(1, mc.get_h2eff(mo_coeff), mc.ncas)
         E_c = np.tensordot(aeri, cascm2, axes=4) / 2
         log.debug('E_c = %s', E_c)
-    if abs(hyb_x) > 1e-10 or abs(hyb_c) > 1e-10:
-        log.debug(('Adding %s * %s CAS exchange, %s * %s CAS correlation to '
-                   'E_ot'), hyb_x, E_x, hyb_c, E_c)
```
MatthewRHermes, line 175: I think that at verbose >= DEBUG level we ought to log the `hyb_x` and `hyb_c` parameters somewhere in this function. I get that this statement is physically incorrect but it is _mathematically_ correct in that it accurately tells the user what the function did. Can we include the hybridizations parenthetically in lines 170 and 178?

pyscf/mcpdft/mcpdft.py
matthew-hennefarth: Thanks!

pyscf/mcpdft/mcpdft.py
```diff
@@ -153,26 +153,32 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
     # (vj_a + vj_b) * (dm_a + dm_b)
     E_j = np.tensordot(vj, dm1) / 2
     # (vk_a * dm_a) + (vk_b * dm_b) Mind the difference!
```
MatthewRHermes, line 155: comment line now meaningless

pyscf/mcpdft/mcpdft.py
MatthewRHermes, line 1: Other than the warning, I'm not sure what this does that is substantively less misleading than what it did before.

matthew-hennefarth: I want to remove the terminology of "CAS exchange" and "CAS correlation". That partitioning doesn't exist, especially if it is being printed out. If anything, `E_x` is closer to a HF exchange term, and the `E_c` is a active-space exchange/correlation term.

pyscf/mcpdft/mcpdft.py
```diff
@@ -153,26 +153,32 @@ def energy_mcwfn(mc, mo_coeff=None, ci=None, ot=None, state=0, casdm1s=None,
...
     log.debug('CAS energy decomposition:')
     log.debug('Vnn = %s', Vnn)
     log.debug('Te + Vne = %s', Te_Vne)
     log.debug('E_j = %s', E_j)
-    log.debug('E_x = %s', E_x)
-    E_c = 0
+
+    if abs(hyb_x - hyb_c) > 1e-10:
+        log.warn("exchange and correlation hybridization differ")
+        log.warn("Your energy surfaces may have discontinuities, see https://github.com/pyscf/pyscf-forge/issues/128")
```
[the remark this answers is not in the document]
matthew-hennefarth: I would think we would want to warn users if they are doing something that is not particularly smart to do, and pointing to the issue where the topic is discussed is helpful. I should probably instead just warn about getting unreliable/inaccurate results given that [you said](https://github.com/pyscf/pyscf-forge/issues/128#issuecomment-2725095501)

> these exchange and correlation energies don't have comparable magnitudes to the exchange and correlation energies from DFT

This has nothing to do with gradients. We raise an exception if we try and do gradient with `hyb_x != hyb_c`.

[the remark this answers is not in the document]
MatthewRHermes: valid

MatthewRHermes (approved):

Issue #144, opened 2025-05-09
[FEATURE TRANSFER] Transfer MC-PDFT to the PySCF core modules


MatthewRHermes:
Attempting to roughly follow #135 

---

### Authors
Matthew R Hermes (University of Chicago)
Dayou Zhang (University of Minnesota)
Thais R Scott (University of Chicago)
Matthew R Hennefarth (University of Chicago)
Bhavnesh Jangid (University of Chicago)
Shirong Wang (Fudan University)
Frédéric Chapoton (University of Strasbourg)

### Feature Description
I propose to transfer the entire multiconfiguration pair-density functional theory (MC-PDFT) implementation from PySCF-Forge to the PySCF core modules, excluding dipole-moment features for which the corresponding implementations in other quantum chemistry methods are not currently hosted in the PySCF core modules. MC-PDFT refers to methods in which the total electronic energy is obtained or derived from a functional of the on-top pair density as well as the total electron density, and these densities are in turn obtained from a multiconfigurational self-consistent field (MCSCF) wave function or wave functions of some sort. The on-top pair density is the probability of two electrons existing simultaneously at one point in space.

The non-classical part of the MC-PDFT energy is called the "on-top energy" and is directly analogous to the exchange-correlation energy in standard density functional theory. In practice, the on-top energy functional is always a generalization or a "translation" of an exchange-correlation functional from Kohn-Sham DFT. MC-PDFT furnishes a practical and effective generalization of KS-DFT functionals from single Slater determinants to general multiconfigurational wave functions because the pair density, unlike the spin density, can simultaneously encode both strong electron correlation effects and spin symmetry constraints.

This feature implements theoretical developments discussed in the following publications:
- Original MC-PDFT publication: [*JCTC* **2014**, *10*, 3669]
- Functional development publications:
  - Fully-translated LDA and GGA on-top functionals: [*JCTC* **2015**, *11*, 4077]
  - Translated meta-GGA on-top functionals: [*PNAS* **2024**, *122*, e2419413121]
  - Global hybrid on-top functionals: [*JPCL* **2020**, *11*, 10158] and [*JCTC* **2020**, *16*, 2274]
- Method extension publications:
  - XMS-PDFT: [*Faraday Discuss* **2020**, 224, 348-372]
  - CMS-PDFT: [*JCTC* **2020**, *16*, 7444]
  - L-PDFT: [*JCTC* **2023**, *19*, 3172]
- Gradient publications:
[...]

PR #145, opened 2025-05-12
Linting to pyscf standards

MatthewRHermes:
When I make a PySCF branch with this code in it (#144) it complains unless I clean this up

### DIFF
--- a/pyscf/grad/test/test_diatomic_gradients.py
+++ b/pyscf/grad/test/test_diatomic_gradients.py
@@ -72,9 +72,9 @@ class KnownValues(unittest.TestCase):
         de_ref = [0.226842531, -0.100538192, -0.594129499] 
         # Numerical from this software
         for i in range (3):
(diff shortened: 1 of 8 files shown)


matthew-hennefarth:
It seems the linter in PySCF-forge here doesn't operate on the test files. Probably should fix that so that these types of PR's are necessary in the future.

matthew-hennefarth:
Actually, in PySCF, the linter shouldn't be operating on test files. See in the [`.flake8`](https://github.com/pyscf/pyscf/blob/07f1e2e8f9839964b13e4ee10919d014b42285c5/.flake8#L18) and [`.ruff.toml`](https://github.com/pyscf/pyscf/blob/07f1e2e8f9839964b13e4ee10919d014b42285c5/.ruff.toml#L1-L3).

Or is the yapf linter that is causing the issue?

sunqm (approved):

PR #149, opened 2025-05-16
Small fix for the LPDFT kernel argument

JangidBhavnesh:
LPDFT kernel doesn't take the dump_chk argument. 

Although, by default it's true which is a good thing, but in case if I don't want to save the information on chkfile or I am running the LPDFT based on CASCI object then calculations are crashing. They can be run, if I pass the dump_chk as false, that's not the solution, yet the LPDFT kernel should take the dump_chk argument.

mc = mcpdft.CASCI(mf, 'tpbe', 2, 2)
mc_ms = mc.multi_state ([.5, .5], "lin")
mc_ms.kernel(verbose=4, dump_chk=False)

@matthew-hennefarth


matthew-hennefarth (approved):

matthew-hennefarth:
Out of curiosity, is the `chk` file not working properly for a CASCI object? With PDFT with a CASCI object? or just L-PDFT with a CASCI object?

MatthewRHermes (approved):

PR #165, opened 2025-11-05
L-PDFT Permanent Dipole Moments

helenclifford:
Adding the implementation of L-PDFT analytic permanent dipole moments and a test case that verifies analytic L-PDFT permanent dipole moment  vs. their numeric values

### DIFF
--- a/examples/prop/03-lpdft-dipole_moment.py
+++ b/examples/prop/03-lpdft-dipole_moment.py
@@ -8,8 +8,6 @@ from pyscf import gto, scf, mcpdft
 from pyscf.mcpdft.lpdft import _LPDFT
 from pyscf.prop.dip_moment.lpdft import _LPDFTDipole
 
-_LPDFT.dip_moment = _LPDFTDipole.dip_moment
-
 mol = gto.M(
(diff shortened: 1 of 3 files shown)


examples/prop/03-lpdft-dipole_moment.py
```diff
@@ -0,0 +1,34 @@
...
+
+"""
+Response-theory calculation of the L-PDFT permanent dipole moment
+"""
+
+from pyscf import gto, scf, mcpdft
+from pyscf.mcpdft.lpdft import _LPDFT
+from pyscf.prop.dip_moment.lpdft import _LPDFTDipole
+
+_LPDFT.dip_moment = _LPDFTDipole.dip_moment
```
MatthewRHermes: To be perfectly clear about this: you need line 9, but you do not need line 11. As soon as you import anything from `pyscf.prop.dip_moment.lpdft`, the Python interpreter executes all of the code in that file, as if you had run `python lpdft.py`, and line 146 of `lpdft.py` already does what line 11 here does.

MatthewRHermes (approved):
Thanks!

PR #173, opened 2026-01-28
fix SF TDA/TDDFT fail and enable tests

jeanwsr:
Currently any run of SF TDA/TDDFT fails because there's no `self.max_space`. For a quick fix I comment it since it's not used. Can be added back if someone want to use it.
Also, almost all tests for SF TDA/TDDFT are skipped because no mcfun installed. Now they are enabled.
Also fix the occri tests fail due to recent pyscf changes.

### DIFF
--- a/pyscf/csf_fci/csf.py
+++ b/pyscf/csf_fci/csf.py
@@ -275,28 +275,33 @@ def pspace (fci, h1e, eri, norb, nelec, transformer, hdiag_det=None, hdiag_csf=N
         except AttributeError:
             csf_addr = csf_addr[np.argsort(hdiag_csf[csf_addr])[:npsp]]
 
+
+
     # To build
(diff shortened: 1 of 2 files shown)


jeanwsr:
Can @Haskiy or @hoshishin please help recover the failing tests?

Haskiy:
I understand what you mean. When I wrote the unit tests initially, I kept six decimal places unchanged. This might be due to two reasons: first, the PySCF 2.5.0 version I was using at the time was related to this, and the tests passed; second, there was a bug in the code with the subspace size (i.e., the max_space parameter) when solving the spin flip TDA equations.
My suggestion is that we no longer use the gradients calculated from the excited states and transition vectors obtained via the davidson method as reference values. Instead, we should use the gradients computed from the excited states and transition vectors obtained via the full diagonalization method as reference values. Then, based on the current calculation results, we can uniformly set the number of decimal places to retain in the unit tests.

Issue #191, opened 2026-04-27
Unit tests failing again, most likely due to new PySCF version


MatthewRHermes:
Several `occri` and `trexio` unit tests are now failing since (I'm guessing) the new PySCF version dropped.

MatthewRHermes:
PR #192 is my debug attempt

MatthewRHermes:
There is an actual discrepancy in the behavior of the PWKRCCSD amplitude equations between PySCF 2.12.0 and 2.13.0. Even if I strictly tighten convergence criteria, revert lindep behavior, and load all wave function data other than the amplitudes themselves from checkpoint files, the discrepancy driving the failure of the [last unfixed unit test](https://github.com/pyscf/pyscf-forge/actions/runs/25340259296) persists. @kylebystrom or @sunqm please help!

kylebystrom:
I will try to take a look at this today. Thanks for bringing this to my attention!

MatthewRHermes:
OK so I can make the test pass on my laptop if I absolutely clobber the plane wave basis, setting the cutoffs in lines 57 and 87 to 200, without changing anything else, at the price of it taking 3 times as long to complete. Is that reasonable? I don't have any intuition about the numerical precision of these methods and these thresholds are all over the place.

kylebystrom:
That's great that increasing the PW cutoff fixes the issue, though it's still very strange that the PySCF updates break the test. The point of the second part of the test (after setting ecut_eri on line 57) is to show that precision is mostly preserved in a small PW basis, which drastically speeds up the calculations. So I will look into it a bit more before we settle on that solution.

The commit that breaks the unit test is pyscf/pyscf@8ae29f6867617ac3dc4798371178fe8ae3987d36. I suspect it has to do with the change in how the Ewald correction is handled, but I haven't pinpointed exactly where that comes in or why it only affects CCSD.

kylebystrom:
As an additional debugging note, the correlation energy for GTO CCSD is identical in 2.12 and 2.13. It's the PWKRCCSD result that changes. The SCF energies are the same.

Edit: The SCF energies the same for GTO and PW because the PW is built from the GTO calculation in this case. Between PySCF 2.12 and 2.13, the total SCF energies differ by about 2e-6 Ha. The occupied orbital energies differ by ~3e-6 to 7e-5 Ha for the PW SCF calculation between 2.12 and 2.13, which might be where the discrepancy arises.

MatthewRHermes:
[...]

PR #192, opened 2026-04-30
Debug PySCF 2.13.0 (#191)

MatthewRHermes:
- Fix memory passing in `csf_fci` (again)
- Update data format of `mo_coeff` in `trexio`
- Account for potential rank reduction in `occri`
- Finer mesh in some unit tests to restore agreement between different forms of density fitting
- Update test references in `pbc/lno` for density matrix elements that differ between versions by scientifically irrelevant phases

~There are still two unit tests that I cannot make pass without changing either the tolerance of the check or the reference value itself, in `pbc/lno` and `pbc/pwscf`. I wonder if anyone else has a commentary on those.~

### DIFF
--- a/pyscf/csf_fci/csf.py
+++ b/pyscf/csf_fci/csf.py
@@ -566,6 +566,9 @@ class FCISolver (CSFFCISolver, direct_spin1.FCISolver):
         return get_init_guess (norb, nelec, nroots, hdiag_csf, self.transformer)
 
     def kernel(self, h1e, eri, norb, nelec, ci0=None, **kwargs):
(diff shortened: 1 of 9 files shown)


MatthewRHermes:
I convinced myself that the failing test in `pbc/lno` was due to irrelevant phase convention differences between PySCF versions. However, the last test, in `pbc/pwscf`, is spooky to me even though it's only slightly off from 1e-6 just because it doesn't change even if I upload v2.12.0 `mo_coeff` data into the unittest.

MatthewRHermes:
#193 fixed the last issue I couldn’t account for, thanks @kylebystrom!!

sunqm (approved):

PR #196, opened 2026-07-09
add double hybrid module

jeanwsr:
Basically developed based on [ajz34/dh](https://github.com/ajz34/dh) while incorporates some feature from [another dh](https://gitee.com/ajz34/pyscf-forge/tree/pre-0.2/pyscf/dh). Address #52.

Capabilities: 
- energy, dipole, polar, grad of DF DH. No non-DF support. No MGGA-DH support.
- D3/D4 support by pyscf-dispersion.

The DH functional parsing is done in dh.xccode, which uses a syntax like pyscf's parse_xc, but with more components like MP2, DFTD3. See examples for more details.

The MP2 step can be done by 3 backends: dfmp2/dfmp2_native (from pyscf, energy only), mp2_ajz (built-in of original dh program, support polar/grad).

cc: @ajz34


examples/dh/01-init.py
```diff
@@ -0,0 +1,51 @@
...
+
+# 3. to_dh — reuse when SCF matches 
+mf_ks = dft.KS(mol, xc="B3LYPg").density_fit().run()
+mf = to_dh(mf_ks, xc="XYG3").run()
+print(f"to_dh reuse XYG3: {mf.e_tot:.8f}")
+
+# to_dh — auto-convert when SCF mismatches 
+mf_ks = dft.KS(mol, xc="PBE0").density_fit().run()
+mf = to_dh(mf_ks, xc="B2PLYP").run()
+print(f"to_dh conv PBE0 -> B2PLYP: {mf.e_tot:.8f}")
```
MatthewRHermes: What exactly is the difference between these three examples?

examples/dh/01-init.py
MatthewRHermes, line 1: Can we put literature citations in this file?

pyscf/dh/dh.py
```diff
@@ -0,0 +1,277 @@
...
+        self.mo_coeff = NotImplemented
+        self.mo_energy = NotImplemented
+        self.mo_occ = NotImplemented
+        self.Co = self.Cv = NotImplemented
+        self.eo = self.ev = NotImplemented
+        self.D = NotImplemented
+        self.nmo = self.nvir = NotImplemented
+        self.so = self.sv = self.sa = NotImplemented
+        self.e_tot = NotImplemented
+        self.eng_tot = self.eng_nc = self.eng_pt2 = self.eng_nuc = self.eng_os = self.eng_ss = NotImplemented
```
MatthewRHermes: This is the first time I've seen `NotImplemented` being used for default values of data members like this. Is there a reason to use `NotImplemented` instead of `None`?

pyscf/dh/grad/dfdh.py
```diff
@@ -0,0 +1,499 @@
...
+    mf_dh.prepare_S_1()
+    mf_dh.prepare_integral()
+    mf_dh.prepare_xc_kernel()
+    mf_dh.prepare_pt2(dump_t_ijab=dump_t_ijab)
+    mf_dh.prepare_lagrangian(gen_W=True)
+    mf_dh.prepare_D_r()
+    mf_dh.prepare_gradient_jk()
+    mf_dh.prepare_gradient_gga()
+    mf_dh.prepare_gradient_pt2()
+    mf_dh.prepare_gradient_enfunc()
```
MatthewRHermes: many scary impure functions

examples/dh/01-init.py
jeanwsr: The DFDH init accepts mf_or_mol, but it's conservative for accepting mf. It refuses continue when SCF mismatches (in xc, etc.). `to_dh` provides a general way which allows SCF mismatch then re-run SCF. (the last one). The second one, to_dh when SCF matches, does not differ from the first one. It's just showing to_dh can handle both condition.

examples/dh/01-init.py
jeanwsr: Added.

pyscf/dh/dh.py
jeanwsr: Personally I also never use this. It's from [ajz34/dh](https://github.com/ajz34/dh). Now replaced by `None`.

pyscf/dh/grad/dfdh.py
jeanwsr: I have managed to rewrite most of them as pure function.

examples/dh/01-init.py
MatthewRHermes: Can you explain this more completely in commentary?

jeanwsr: Updated.

MatthewRHermes:
I know it's overwhelming dealing with the spurious test failures, but it looks like one or two of the failures in your CI run are actually your own stuff, if I'm not misreading.

jeanwsr:
> I know it's overwhelming dealing with the spurious test failures, but it looks like one or two of the failures in your CI run are actually your own stuff, if I'm not misreading.

I pulled master, now everything is fine.

MatthewRHermes (approved):

PR #201, opened 2026-07-25
PV contributions to energy

juanjoaucar:
Allow PV energy contributions to be computed from density matrices


juanjoaucar:
Dear PySCF developers,

I don't understand why Lint fails, in files that I didn't modify.

MatthewRHermes:
> Dear PySCF developers,
> 
> I don't understand why Lint fails, in files that I didn't modify.

recent ruff update is now flagging a million new things. it's not your fault.

pyscf/pv/energy.py
```diff
@@ -117,18 +162,32 @@ def fc_expval(mol, mf, atom):
...
+
         expval_perorb[k] = expval_LL_k + expval_SS_k
 
+
     return fac * expval_perorb.real
 
 
 
-def Epv_atom(mol, mf, atom_index):
+def Epv_atom(mol, mf, atom_index, dm=None):
```
MatthewRHermes, line 175: So if I'm reading this right, Epv is calculated in one way, then if `dm` is provided that result is completely discarded and it's calculated from scratch in a completely different way. Why? That's confusing. Why not just branch on whether `dm` is provided right away? Or just make two different functions?

juanjoaucar: @MatthewRHermes thanks!

I just did wrong. I modified it according to your suggestion in 1069d6b78af41a35e2a98b1352329b0dc008df27.

I'm just willing to allow to obtain these quantities by importing the WF from previous calculations or other codes

sunqm (approved):

PR #200, opened 2026-07-23
Pull Request for Multireference Methods Using Nonorthogonal Orbitals: GBCI, GB-PDFT, and XMS-GB-PDFT. 

Ohhminseok:
M. Oh, J. Park, B. Kim, H. Lim, and S. Lee, Orbital-Relaxed Bath Theory for Charge-Transfer Processes in Transition-Metal Complexes, PCCP, 28, 13294 (2026)

### Overview

This PR introduces grouped-bath configuration interaction (GBCI) and
GBCI-based pair-density functional theory methods to PySCF-forge.

The previous `pyscf.sfnoci` implementation has been renamed and reorganized as
`pyscf.gbci`. In addition to the namespace change, routines that were previously
embedded in `sfnoci.py` and `direct_sfnoci.py` have been separated into reusable
modules, and new GBPDFT, MS-GBPDFT, and XMS-GBPDFT drivers are added.

---

#### 1. GBCI core (`gbci/gbci.py`, `gbci/direct_gbci.py`)

Renames and reorganizes the previous SFNOCI implementation as GBCI:

- renames the user-facing `sfnoci.py` workflow to `gbci.py`
- renames `direct_sfnoci.py` to `direct_gbci.py`
- keeps the grouped-bath CI solver structure
- supports molecular-orbital, atom-based, and occupation-based grouping
- updates the C contraction library naming from `libsfnoci` to `libgbci`

#### 2. Frozen active-space SCF utilities (`gbci/fasscf.py`)

Separates the frozen active-space SCF routines previously embedded in
`sfnoci.py` into a standalone reusable module.

This module is used by the GBCI orbital-relaxation workflow and includes:

- frozen active-space SCF updates
- relaxation of the complementary orbital space
- a Newton-based optimization path for the orbital-relaxed bath procedure

#### 3. Density-matrix utilities (`gbci/rdm.py`)

Separates and extends the density-matrix routines previously embedded in
`direct_sfnoci.py`.

The previous state-specific spin-separated 1-RDM and 2-RDM routines were
refactored and optimized. This PR also adds transition 1-RDM construction
between GBCI states and exposes reusable intermediates for GBPDFT and
XMS-GBPDFT calculations.

These intermediates are important because GBPDFT avoids explicitly constructing
full 2-RDM/2-CDM objects when evaluating the translated on-top pair density.

#### 4. GBPDFT and on-top pair density (`gbci/gbpdft.py`, `gbci/otpd.py`)

Adds GBCI-reference pair-density functional theory:

[...]

PR #203, opened 2026-07-27
Add generalized active space CI solver with OpenMP C kernels

yideng2025:
## Summary

This PR adds a determinant-based generalized active space configuration interaction (GASCI) implementation to `pyscf-forge`.

It includes:

* OpenMP-enabled C kernels for GAS determinant-space construction, Hamiltonian contraction, Hamiltonian diagonal construction, and reduced density matrices.
* Python bindings and an `FCISolver` interface for the GAS FCI backend.
* A `GASCI` driver following the existing PySCF CASCI interface.
* Support for explicit spin-supergroup, supergroup, cumulative-occupation, and RAS-style restrictions.
* Unit tests and examples covering the main GASCI workflows.

### C backend

The new GAS FCI backend constructs the allowed determinant space from legal alpha/beta occupation sectors and organizes determinants into GAS blocks.

The backend provides:

* GAS determinant-string and block construction.
* Compressed single-excitation link tables.
* OpenMP-parallel Hamiltonian contraction.
* Hamiltonian diagonal construction.
* Spin-resolved one- and two-particle RDMs.
* Transition RDMs.
* Reusable contraction and RDM plans.

The C backend is built as `libfci_gas` through the CMake configuration in `pyscf/lib`.

### Python interface

The Python layer provides:

* Normalization and validation of GAS restrictions.
* C-library loading and data-structure management.
* `pyscf.mcscf.fci_gas.FCISolver`.
* `pyscf.mcscf.gasci.GASCI`.
* Single-root and multiroot calculations.
* State-averaged (for future GASSCF).
* Spin-square evaluation and spin-penalty support.
* Consistent propagation of `FCISolver.spin` when inferring active alpha/beta electron counts.
* State and transition RDM interfaces.
* Natural-orbital and GAS pseudo-natural-orbital analysis.
* GAS determinant-space diagnostics.

A GAS containing the complete active-space determinant set is tested against the existing PySCF CAS backend.

## GAS restriction formats

The interface supports four equivalent or complementary ways to define a GAS space:

* `spin-supergroup`: explicit alpha and beta occupations in every GAS subspace.
* `supergroup`: explicit total occupations in every GAS subspace.
* `cumulative-occ`: lower and upper bounds on cumulative subspace occupations.
* `ras`: RAS1 hole and RAS3 electron restrictions.

The restrictions are normalized to the explicit spin-supergroups used by the C backend.

## Tests

The following validation was performed:

* Dedicated GCC 13.2.0 and CMake 3.26 OpenMP build: passed.
* GAS backend smoke test: passed.
* Extended Python-layer regression test: passed.
* Formal GASCI unit tests: 10 tests passed.
* GASCI examples 90, 91, and 92: passed.
* Source and whitespace audit with `git diff --check`: passed.

The formal unit tests can be run with:

```bash
python -m unittest discover \
    -v \
    -s pyscf/mcscf/test \
    -p 'test_gasci.py'
```

The included examples demonstrate:

* GASCI with explicit spin-supergroups.
* Equivalent GAS restriction formats.
* Multiroot GASCI with RDM, transition-RDM, spin, and natural-orbital analysis.

## Scope

This PR is intentionally limited to GASCI.

Orbital-optimized GASSCF functionality is not included because it requires additional implementation and independent correctness validation.


jeanwsr:
Might be better to add a CMakeLists.txt under lib/mcscf, like the one of pwscf.

yideng2025:
> Might be better to add a CMakeLists.txt under lib/mcscf, like the one of pwscf.

Thanks for the suggestion. Addressed in 4bb49e3: the GAS FCI target is now defined in pyscf/lib/mcscf/CMakeLists.txt, while the top-level CMake file only handles the build-mode routing and add_subdirectory(mcscf).

The Copilot comments were also addressed by making the binomial-table initialization thread-safe and resolving diagonal cutoff ties deterministically with a regression test.

Validation completed locally: GCC 13.2.0/CMake 3.26 OpenMP build, 11 unit tests, backend and Python-layer regression tests, 20 × 32-thread initialization stress tests, and examples 90–92 all passed.

pyscf/mcscf/test/test_gasci.py
matthew-hennefarth, line 1: Would be nice to check the energy value you get here to the values from OpenMolcas for example (where GAS has been implemented previously). At the least, you should also make sure the energy of the GASCI calculation does not change, most of the tests in the file are just checking the RDM elements (and specifically trace of the RDM).

yideng2025: > Would be nice to check the energy value you get here to the values from OpenMolcas for example (where GAS has been implemented previously). At the least, you should also make sure the energy of the GASCI calculation does not change, most of the tests in the file are just checking the RDM elements (and specifically trace of the RDM).

Thank you Matthew. I have addressed both points.

(1) I added fixed-orbital GASCI validation examples against OpenMolcas for six systems: NO+, NO, O2, O2+, O2-, and AlFe2O4+. The PySCF calculations use the same molecular definition and MO coefficients as the corresponding OpenMolcas calculations. I did not include the original OpenMolcas `.h5` or `.out` output files in the repository; instead, I saved the required molecular data and MO coefficients as `.npz` files under `data/`, together with the corresponding `.xyz` geometries.

The PySCF GASCI energies are compared directly with the OpenMolcas results, and all tested roots agree within 1E-8 Eh. This tolerance is consistent with the fact that OpenMolcas prints the total energies to eight digits after the decimal point.

(2) I also added an explicit O2 triplet energy regression test to `pyscf/mcscf/test/test_gasci.py`. It checks the GASCI ground state energy against the fixed reference value `-147.5163002382864 Eh`. This should catch future changes in the GASCI energy, rather than only checking quantities such as RDM traces.

The updated GASCI test suite now passes all 13 tests.

matthew-hennefarth: The examples aren't actually run in any CI/CD pipeline, so they can drift undetected. Examples should not be the core validation, but just demonstration of how to use the code. I would prefer at least 1 (lightweight) test that validates against the OpenMolcas result, even if there is slightly looser tolerance (like 1e-6 or so). This should ideally be in addition to the o2 triplet test.

pyscf/mcscf/test/test_gasci.py
```diff
@@ -235,6 +235,53 @@ def setUpClass(cls):
...
+        mc.fcisolver.nroots = 1
+        mc.fcisolver.max_cycle = 300
+        mc.fcisolver.conv_tol = 1e-12
+        mc.kernel()
+        self.assertTrue(mc.converged)
+
+        numpy.testing.assert_allclose(
+            mc.e_tot,
+            -147.5163002382864,
+            atol=1e-9,
```
matthew-hennefarth: Should be fine, but down the line might be too tight. Could always loosen to 1e-7 (which is a value used in the mcscf module of pyscf).

pyscf/mcscf/test/test_gasci.py
yideng2025: > The examples aren't actually run in any CI/CD pipeline, so they can drift undetected. Examples should not be the core validation, but just demonstration of how to use the code. I would prefer at least 1 (lightweight) test that validates against the OpenMolcas result, even if there is slightly looser tolerance (like 1e-6 or so). This should ideally be in addition to the o2 triplet test.

That makes sense. I moved the core OpenMolcas validation into `test_gasci.py` as a CI-run regression test. The new test embeds the exact NO+ molecular definition and fixed MO coefficients directly in the test file, checks MO orthonormality, computes the lowest 30 states, filters the 10 singlet roots, and compares all of them against the OpenMolcas reference energies with `atol=1e-7`.

I also simplified the examples so that they serve as demonstrations rather than the main validation suite. They now contain one O2 multiroot example and one AlFe2O4+ single-root example, with representative output included as comments.

pyscf/mcscf/test/test_gasci.py
yideng2025: > Should be fine, but down the line might be too tight. Could always loosen to 1e-7 (which is a value used in the mcscf module of pyscf).

Agreed. I relaxed the O2 triplet energy regression tolerance from `1e-9` to `1e-7`, with `rtol=0`. The new NO+/OpenMolcas multiroot regression also uses an absolute tolerance of `1e-7`.

pyscf/lib/mcscf/fci_contract_gas.c
```diff
@@ -0,0 +1,2609 @@
...
+static inline double abba_gvalue(const double *restrict g,
+                                 uint32_t nnorb, uint16_t opb, uint16_t opa)
+{
+        return g[(uint64_t)opb * nnorb + opa];
+}
+
+static double *build_abba_gos(const double *restrict eri, uint32_t nnorb)
+{
+        size_t n = (size_t)nnorb * (size_t)nnorb;
+        double *gos = malloc(n * sizeof(*gos));
```
MatthewRHermes: I struggle to convince myself that this is safe from memory leaks. Can you not do this in the python layer?

MatthewRHermes:
Merge conflicts created by #200, please resolve

pyscf/lib/mcscf/fci_rdm_gas.c
```diff
@@ -0,0 +1,2407 @@
...
+                plan, bra, ket, dm1, dm1b, dm2aa, dm2ab, dm2bb);
+        if (status == GAS_SUCCESS) {
+                for (size_t i = 0; i < n1; i++) dm1[i] += dm1b[i];
+                for (uint32_t p = 0; p < norb; p++) {
+                        for (uint32_t q = 0; q < norb; q++) {
+                                for (uint32_t r = 0; r < norb; r++) {
+                                        for (uint32_t s = 0; s < norb; s++) {
+                                                size_t i = rdm2_index(p, q, r, s, norb);
+                                                size_t j = rdm2_index(r, s, p, q, norb);
+                                                dm2[i] = dm2aa[i] + dm2bb[i] +
```
MatthewRHermes: IMO this belongs in the Python layer

pyscf/mcscf/test/test_gasci.py
```diff
@@ -0,0 +1,723 @@
...
+    for pq in range(npair):
+        for rs in range(npair):
+            lo, hi = min(pq, rs), max(pq, rs)
+            eri[pq, rs] = 0.011 * (lo + 1) + 0.003 * (hi + 2)
+    return h1e, eri
+
+
+class TestGASRestrictions(unittest.TestCase):
+
+    def test_nelec_spin_inference(self):
```
MatthewRHermes, line 61: Is this logic unique to GAS as opposed to CAS? Does `pyscf.mcscf.addons._unpack_nelec` do the same thing as `addons_gas.as_nelec_tuple`?

pyscf/mcscf/test/test_gasci.py
```diff
@@ -0,0 +1,723 @@
...
+        ras = {"max_holes": 1, "max_particles": 1}
+        cumulative = [[1, 2], [3, 4], [4, 4]]
+
+        kernel_orbs, ras_blocks = addons_gas.normalize_gas_spec(
+            gas_orbs, nelec, ras, "ras")
+        cumulative_blocks = addons_gas.normalize_gas_restr(
+            gas_orbs, nelec, cumulative, "cumulative-occ")
+
+        self.assertEqual(kernel_orbs, gas_orbs)
+        numpy.testing.assert_array_equal(ras_blocks, cumulative_blocks)
```
MatthewRHermes, line 159: Are these two tests thorough enough to guarantee no bug for any choice of GAS or RAS?

pyscf/mcscf/addons_gas.py
```diff
@@ -0,0 +1,922 @@
...
+            "occupations and a final bound equal to the active electrons."),
+        GAS_RESTR_RAS: (
+            "gas_restr_type='ras': gas_orbs has exactly three entries "
+            "(RAS1, RAS2, RAS3), which may be zero, and gas_restr is exactly "
+            "{'max_holes': integer, 'max_particles': integer}."),
+    }
+    numeric = (
+        "All occupations, bounds, orbital counts, and electron counts must "
+        "be exact integers; integral values such as 4.0 are accepted, but "
+        "non-integral, Boolean, non-finite, and negative values are rejected.")
```
MatthewRHermes: Do we really need to support "integral values such as 4.0"? Can't we just demand integers and use `pyscf.lib.isinteger` to check?

pyscf/mcscf/addons_gas.py
```diff
@@ -0,0 +1,922 @@
...
+        (StateAverageGASCI, mc.__class__))
+
+
+def state_average_(mc, weights=(0.5, 0.5), wfnsym=None):
+    """In-place version of :func:`state_average`."""
+
+    weighted = state_average(mc, weights, wfnsym)
+    mc.__class__ = weighted.__class__
+    mc.__dict__ = weighted.__dict__
+    return mc
```
MatthewRHermes, line 282: What is the use case for these features at this stage? Obviously you need them when we come to GASSCF, but is there some functionality that this PR enables that actually uses these things?

pyscf/mcscf/addons_gas.py
```diff
@@ -0,0 +1,922 @@
...
+    return numpy.ascontiguousarray(arr, dtype=numpy.int32)
+
+
+def _integer_scalar(value, label):
+    """Return one exact non-boolean integer scalar."""
+
+    arr = _exact_integer_array(value, label)
+    if arr.ndim != 0:
+        raise ValueError("%s must be one integer scalar" % label)
+    return int(arr)
```
MatthewRHermes: This seems overengineered. Maybe it made sense as a debugging tool but are we sure we need it?

pyscf/mcscf/addons_gas.py
```diff
@@ -0,0 +1,922 @@
...
+
+
+def cas_blocks(norb, nelec):
+    na, nb = as_nelec_tuple(nelec)
+    return numpy.array([[na, nb]], dtype=numpy.int32)
```
MatthewRHermes: So in other words, `numpy.array (as_nelec_tuple (nelec), dtype=numpy.int32)[None,:]`. Does this need to be its own function?

pyscf/mcscf/addons_gas.py
```diff
@@ -0,0 +1,922 @@
...
+
+
+def spin_supergroup_blocks(gas_orbs, nelec, rows):
+    """Validate and canonicalize an explicit legal block set D."""
+
+    gas_orbs = _integer_vector(gas_orbs, "gas_orbs")
+    blocks = normalize_blocks(rows, len(gas_orbs))
+    check_kernel_limits(gas_orbs, nelec, blocks)
+    return blocks
```
MatthewRHermes: Function is never called by anything

pyscf/mcscf/fci_gas.py
```diff
@@ -0,0 +1,1212 @@
...
+    if count == hdiag.size:
+        selected = numpy.arange(hdiag.size, dtype=numpy.intp)
+    else:
+        cutoff = numpy.partition(hdiag, count - 1)[count - 1]
+        selected = numpy.flatnonzero(hdiag < cutoff)
+        ntied = count - selected.size
+        tied = numpy.flatnonzero(hdiag == cutoff)[:ntied]
+        selected = numpy.concatenate((selected, tied))
+    order = numpy.lexsort((selected, hdiag[selected]))
+    return numpy.asarray(selected[order], dtype=numpy.intp)
```
MatthewRHermes: Does this solve a real problem that needs to be solved? For FCI, PySCF just does this:
```
    if hdiag.size <= np:
        addr = numpy.arange(hdiag.size)
    else:
        try:
            addr = numpy.argpartition(hdiag, np-1)[:np].copy()
        except AttributeError:
            addr = numpy.argsort(hdiag)[:np].copy()
``` 
Is that not fit for purpose in the GASCI case? The likelihood of a bitwise-perfect tie between 64-bit floating-point numbers seems remote to me, even for high-symmetry molecules, because any OpenMP process at any point in the stack will introduce roundoff nondeterminism.

pyscf/mcscf/fci_gas.py
```diff
@@ -0,0 +1,1212 @@
...
+            linear_spin_penalty = False
+            if spin_penalty is not None:
+                (spin_shift, spin_target, spin_minimum,
+                 spin_eigenvalues) = spin_penalty
+                linear_spin_penalty = spin_target < spin_minimum + 0.1
+
+            plan = ctypes.c_void_p()
+            status = self.lib.fci_contract_gas_plan_create(
+                ctypes.byref(plan), gas.c_ptr, _gaslib.double_ptr(h2e))
+            _check_status(status, "fci_contract_gas_plan_create")
```
MatthewRHermes: It's a little odd to me that the two "plan" objects (Hamiltonian and S^2) do not appear to be related at all in the Python layer. One of them is a class and the other is a weird C object.

pyscf/mcscf/fci_gas.py
```diff
@@ -0,0 +1,1212 @@
...
+                        exploration_guess = []
+                        for candidate in guess[:max(1, nroots)]:
+                            norm = numpy.linalg.norm(candidate)
+                            if norm > 0.0:
+                                exploration_guess.append(candidate / norm)
+                        guess = projected_guess + exploration_guess
+                        self.spin_penalty_method = (
+                            "projected-plus-global-davidson")
+                    if nroots == 1 and len(guess) == 1:
+                        guess = guess[0]
```
MatthewRHermes, line 1148: Can all of this be collapsed into a modified version of `get_init_guess` that takes an optional `gas` kwarg?

pyscf/mcscf/fci_gas.py
```diff
@@ -0,0 +1,1212 @@
...
+                        if linear_spin_penalty:
+                            ss_vector -= spin_target * vec
+                            result += spin_shift * ss_vector
+                            return result
+                        tmp = ss_vector - spin_target * vec
+                        correction = numpy.asarray(
+                            spin_plan.contract(tmp)).reshape(-1)
+                        correction -= spin_target * tmp
+                        result += spin_shift * correction
+                        return result
```
MatthewRHermes, line 1166: I wish you used `self.contract_2e` in here. As it stands, you have what amount to two different implementations of `contract_2e`, one of which is hidden in this code, and that is a potential source of future technical debt.

pyscf/mcscf/gasci.py
```diff
@@ -0,0 +1,1157 @@
...
+MO_ORTH_WARN_TOL = getattr(
+    __config__, "mcscf_gasci_mo_orth_warn_tol", 1e-7)
+MO_ORTH_ERROR_TOL = getattr(
+    __config__, "mcscf_gasci_mo_orth_error_tol", 1e-5)
+
+
+def h1e_for_gas(mc, mo_coeff=None, ncas=None, ncore=None):
+    """Return the effective one-electron Hamiltonian and core energy."""
+
+    return casci.h1e_for_cas(mc, mo_coeff, ncas, ncore)
```
MatthewRHermes: `h1e_for_gas = casci.h1e_for_cas`?

MatthewRHermes (changes requested):
Some final thoughts, after reviewing and reflecting on my own comments. I sense three overall buckets of concern:

1. Do we need X? Doesn't PySCF already have a function for X?
2. Let's please do everything possible to guarantee memory safety. My preference is to allocate everything possible in the Python layer, and when I can't do that, I always try to make sure the `alloc` and the `free` command for any given thing appear in the same function, and I try to keep them as close to each other as possible.
3. Some cleanup may be possible in the `fci_gas.py` `kernel` function.

To this I will here add:

4. Can we get an example input file for a simple RASCI calculation? Say for a 3d transition metal active space with semicore p orbitals in RAS1? Of course it's possible to infer how to do it by studying the present examples very closely but users need their hands held.

I'm willing to be told "no" about most of my comments, but please at least pull master and resolve merge conflicts created by #200.

pyscf/mcscf/test/test_gasci.py
MatthewRHermes, line 1: OK, sorry, I lied, one more thing before I'm done. There is no specific test here of `contract_2e` or `contract_spin` as such. If you had a tool to project CI vectors back and forth between the GAS Hilbert space and the FCI Hilbert space, this would be easy to implement even for complicated GAS setups (i.e., not just the GAS-as-CAS limiting case). The GAS Hilbert space is entirely within the FCI Hilbert space, so you can insert resolutions of the identity as you please,

$$
\langle I_{\textrm{GAS}}|\hat{O}|\Psi_{\textrm{GAS}}\rangle = \sum_{J,K} \langle I_{\textrm{GAS}}|J_{\textrm{FCI}}\rangle\langle J_{\textrm{FCI}}|\hat{O}|K_{\textrm{FCI}}\rangle\langle K_{\textrm{FCI}}|\Psi_{\textrm{GAS}}\rangle
$$

which implies that if you
1. Map a GAS CI vector `x` (i.e., packed flat ndarray) into a FCI vector `c` (rectangular ndarray with many zeros)
2. Use the FCI `contract_` function: `oc = contract_o (c)`
3. Project the operator-vector product `oc` back into the GAS space (necessarily discarding some nonzero coefficients corresponding to illegal determinants)

then the outcome should be the same as if you did the contract operation with the GAS function: `ox = contract_o (x)`.

MatthewRHermes: N.B. `gas2fci`/`fci2gas` converter functions for the CI vector would also be very useful for developing compatibility with the CSF solver.

yideng2025: > OK, sorry, I lied, one more thing before I'm done. There is no specific test here of `contract_2e` or `contract_spin` as such. If you had a tool to project CI vectors back and forth between the GAS Hilbert space and the FCI Hilbert space, this would be easy to implement even for complicated GAS setups (i.e., not just the GAS-as-CAS limiting case). The GAS Hilbert space is entirely within the FCI Hilbert space, so you can insert resolutions of the identity as you please,
> 
> ⟨ I GAS | O ^ | Ψ GAS ⟩ = ∑ J , K ⟨ I GAS | J FCI ⟩ ⟨ J FCI | O ^ | K FCI ⟩ ⟨ K FCI | Ψ GAS ⟩
> 
> which implies that if you
> 
> 1. Map a GAS CI vector `x` (i.e., packed flat ndarray) into a FCI vector `c` (rectangular ndarray with many zeros)
> 2. Use the FCI `contract_` function: `oc = contract_o (c)`
> 3. Project the operator-vector product `oc` back into the GAS space (necessarily discarding some nonzero coefficients corresponding to illegal determinants)
> 
> then the outcome should be the same as if you did the contract operation with the GAS function: `ox = contract_o (x)`.

Hi Matt. Added reusable `gas2fci` / `fci2gas` converters and operator-level projection tests in `83e2ba7`.

The tests compare the GAS contractions against the corresponding full-FCI contractions followed by projection, i.e. $O_{\mathrm{GAS}}x = P O_{\mathrm{FCI}} P^\dagger x$, for explicit spin-supergroup, supergroup, and RAS restrictions.

For the spin-complete supergroup and RAS cases, the spin-square tests also verify that the GAS space is invariant under $\hat{S}^2$, i.e. $\hat{S}^2_{\mathrm{FCI}} P^\dagger x = P^\dagger \hat{S}^2_{\mathrm{GAS}} x$.

Thanks for the suggestion — the converters should also be useful for future CSF-solver compatibility.

pyscf/mcscf/fci_gas.py
yideng2025: > It's a little odd to me that the two "plan" objects (Hamiltonian and S^2) do not appear to be related at all in the Python layer. One of them is a class and the other is a weird C object.

Addressed this in `be5ca13`. I added a Python `GasContractPlan` wrapper around the reusable C Hamiltonian contraction plan, so the Hamiltonian plan now has explicit Python-side ownership and lifetime management rather than exposing a raw `ctypes.c_void_p` in the solver code.

The plan retains both the `GasSpace` reference and the normalized ERI array for the full C-plan lifetime, and provides the same context-manager-style lifetime handling as the existing RDM plan.

pyscf/mcscf/fci_gas.py
yideng2025: > I wish you used `self.contract_2e` in here. As it stands, you have what amount to two different implementations of `contract_2e`, one of which is hidden in this code, and that is a potential source of future technical debt.

Also addressed this in `be5ca13`. The Davidson path now goes through `self.contract_2e` rather than a separate private contraction implementation.

`contract_2e` can take an existing `GasContractPlan`, so the Davidson solver still creates the contraction plan only once and reuses it for all matvecs. The ordinary public call creates a temporary plan internally. This keeps a single Python contraction path without giving up the reusable-plan performance.

I also added a regression test that reuses one `GasContractPlan` for multiple CI vectors and checks the results against the ordinary `contract_2e` path.

pyscf/lib/mcscf/fci_rdm_gas.c
yideng2025: > IMO this belongs in the Python layer

Addressed in `8fa10f9`. The spin-free RDM assembly is now done entirely in the Python layer. The C kernel only computes the spin-resolved RDM components through the reusable RDM plan (`make_rdm1s` / `make_rdm12s`), while `GasRDMPlan.make_rdm1` and `make_rdm12` assemble the spin-free RDMs with NumPy.

I also removed the now-unused one-shot and spin-free C RDM entry points, so there is only one C RDM execution path. A regression test directly checks the Python spin-free assembly against the spin-resolved components.

pyscf/lib/mcscf/fci_contract_gas.c
yideng2025: > I struggle to convince myself that this is safe from memory leaks. Can you not do this in the python layer?

Addressed in `2b1443c`. I do not think the previous implementation had an actual memory leak: the C-allocated `gos` buffer was freed in the one-shot path and by `fci_contract_gas_plan_free` for the reusable-plan path. However, I agree that the ownership was harder to reason about than necessary.

`gos` is now constructed and owned by the Python `GasContractPlan`, alongside the normalized ERI array. The C plan only borrows pointers to the GAS space, ERI, and `gos`, while owning only its execution workspaces and OpenMP task data. This makes the lifetime explicit: the Python object that owns the contraction data is also responsible for releasing it.

I also removed the now-unused one-shot `fci_contract_gas_2e` entry point and its separate C-side allocation/free path, leaving a single reusable contraction-plan path.

pyscf/mcscf/test/test_gasci.py
yideng2025: > Is this logic unique to GAS as opposed to CAS? Does `pyscf.mcscf.addons._unpack_nelec` do the same thing as `addons_gas.as_nelec_tuple`?

Addressed in `dded087`. This electron-count logic is not GAS-specific. I removed `addons_gas.as_nelec_tuple` and now use `pyscf.fci.addons._unpack_nelec` throughout the GAS solver and GASCI driver, so the alpha/beta electron inference follows the existing PySCF FCI convention directly. The corresponding test now checks that the GAS solver uses the same unpacking behavior as the PySCF helper.

pyscf/mcscf/addons_gas.py
yideng2025: > Do we really need to support "integral values such as 4.0"? Can't we just demand integers and use `pyscf.lib.isinteger` to check?

Addressed in `9569800`. I agree that supporting integral floating-point values such as `4.0` is unnecessary here. GAS restriction inputs now require actual integer values, using `pyscf.lib.isinteger` for validation, and floating-point and Boolean values are rejected.

I also added regression coverage for orbital counts, supergroup/cumulative occupations, and RAS limits, while retaining support for standard NumPy integer arrays.

pyscf/mcscf/addons_gas.py
yideng2025: > What is the use case for these features at this stage? Obviously you need them when we come to GASSCF, but is there some functionality that this PR enables that actually uses these things?

Yes. There is already a GASCI-level use case for these features, independent of orbital optimization.

For a fixed-orbital multiroot GASCI calculation, the state-average wrapper provides weighted multistate energies and RDMs while retaining the individual root energies and root-resolved RDM/spin interfaces. The current GASCI tests exercise this directly by checking the weighted total energy and weighted 1-RDM against the corresponding sums over the individual roots. So I would prefer to keep the state-average support in this PR.

pyscf/mcscf/addons_gas.py
yideng2025: > This seems overengineered. Maybe it made sense as a debugging tool but are we sure we need it?

Addressed in `e928f40`. I agree that `_integer_scalar` was unnecessary abstraction for the few scalar inputs here. I removed it and now validate `max_holes`, `max_particles`, and `base` directly with `pyscf.lib.isinteger` before converting them to Python integers.

Array and vector inputs still go through the existing array validation helpers, while scalar validation is now kept local to the corresponding arguments. I also added regression coverage for invalid floating-point and Boolean `base` values.

yideng2025:
> Merge conflicts created by #200, please resolve

Addressed in `3b845a4`.

pyscf/mcscf/addons_gas.py
yideng2025: > So in other words, `numpy.array (as_nelec_tuple (nelec), dtype=numpy.int32)[None,:]`. Does this need to be its own function?

Addressed in `336a2d2`. Agreed. `cas_blocks` was only wrapping construction of the single CAS-limit `(Nalpha, Nbeta)` block and did not justify a separate helper.

I removed it and inlined the block construction at its two call sites. In `GasSpace`, the electron count has already been normalized, so the block is constructed directly from `self.nelec`.

pyscf/mcscf/addons_gas.py
yideng2025: > Function is never called by anything

Addressed in `088d544`. Agreed. `spin_supergroup_blocks` had no callers, so I removed the unused helper.

pyscf/mcscf/fci_gas.py
yideng2025: > Does this solve a real problem that needs to be solved? For FCI, PySCF just does this:
> 
> ```
>     if hdiag.size <= np:
>         addr = numpy.arange(hdiag.size)
>     else:
>         try:
>             addr = numpy.argpartition(hdiag, np-1)[:np].copy()
>         except AttributeError:
>             addr = numpy.argsort(hdiag)[:np].copy()
> ```
> 
> Is that not fit for purpose in the GASCI case? The likelihood of a bitwise-perfect tie between 64-bit floating-point numbers seems remote to me, even for high-symmetry molecules, because any OpenMP process at any point in the stack will introduce roundoff nondeterminism.

Addressed in `93604c7`. Agreed. I removed the GAS-specific deterministic tie-handling logic and now use the same `argpartition`-based diagonal selection strategy as PySCF FCI, with the usual `argsort` fallback.

I kept `_lowest_hdiag_addresses` as a small shared helper because the same selection is used by the initial-guess, pspace, and small-space kernel paths, but it no longer implements any GAS-specific ordering policy. I also removed the tie-ordering regression test.

pyscf/mcscf/fci_gas.py
yideng2025: > Can all of this be collapsed into a modified version of `get_init_guess` that takes an optional `gas` kwarg?

Addressed in `bcdc313`. Yes. I updated `get_init_guess` to take an optional `gas` argument. When an existing `GasSpace` is provided, it now uses `gas.ndet` directly instead of reconstructing the GAS specification through `space_info`.

The solver kernel now passes its already-constructed `GasSpace` into `get_init_guess`, while the public `gas=None` path retains the previous behavior. I kept user-`ci0` handling and spin-projection logic in `kernel`, since those are solver-orchestration concerns rather than default guess construction.

I also added a regression test confirming that the ordinary and reused-`GasSpace` paths generate identical initial guesses.

pyscf/mcscf/gasci.py
yideng2025: > `h1e_for_gas = casci.h1e_for_cas`?

Addressed in `73bede5`. Agreed. `h1e_for_gas` was only a forwarding wrapper around `casci.h1e_for_cas`, so I replaced it with a direct alias: `h1e_for_gas = casci.h1e_for_cas`.

pyscf/mcscf/test/test_gasci.py
yideng2025: > Are these two tests thorough enough to guarantee no bug for any choice of GAS or RAS?

I would not claim that these two tests alone can guarantee correctness for every possible GAS or RAS specification. Their purpose is narrower: they test the Python-layer conversion between the supported user-facing restriction representations.

Internally, all of these representations are normalized to the same canonical set of spin-supergroups before entering the C kernels. `test_equivalent_restriction_types` checks that explicit spin-supergroups, spin-free supergroups, and cumulative-occupation bounds produce the same canonical block set for an equivalent specification, while `test_ras_matches_cumulative_bounds` checks the corresponding RAS-to-cumulative conversion.

The correctness of the resulting GAS Hilbert space and the C operators is tested separately by the GAS↔FCI projection tests added in `83e2ba7`, including nontrivial spin-supergroup, supergroup, and RAS spaces. Those tests compare the GAS Hamiltonian contraction directly against full-FCI contraction followed by projection, and also test the spin operator for spin-complete spaces.

So I view these two tests as checks of the input-conversion layer rather than exhaustive validation of every possible GAS/RAS choice.

yideng2025:
> Some final thoughts, after reviewing and reflecting on my own comments. I sense three overall buckets of concern:
> 
> 1. Do we need X? Doesn't PySCF already have a function for X?
> 2. Let's please do everything possible to guarantee memory safety. My preference is to allocate everything possible in the Python layer, and when I can't do that, I always try to make sure the `alloc` and the `free` command for any given thing appear in the same function, and I try to keep them as close to each other as possible.
> 3. Some cleanup may be possible in the `fci_gas.py` `kernel` function.
> 
> To this I will here add:
> 
> 4. Can we get an example input file for a simple RASCI calculation? Say for a 3d transition metal active space with semicore p orbitals in RAS1? Of course it's possible to infer how to do it by studying the present examples very closely but users need their hands held.
> 
> I'm willing to be told "no" about most of my comments, but please at least pull master and resolve merge conflicts created by #200.

For (4), I added a simple transition metal RASCI example in `f72f9f9`.

The new example uses sextet Al2FeO4+ with

* RAS1 = O 2p (12 orbitals), allowing up to 2 holes
* RAS2 = Fe 3d (5 orbitals)
* RAS3 = Fe 4d (5 orbitals), allowing up to 2 particles

This gives a RASCI(29e,22o) space with 180,486 determinants.

The example is also cross-validated against a fixed-orbital OpenMolcas RASCI calculation. The OpenMolcas and PySCF energies are `-2041.069566720000 Eh` and `-2041.069566719117 Eh`, respectively, differing by about `8.8e-10 Eh`.

yideng2025:
I found that the MO ordering in the previous Al2FeO4+ RASCI example was not preserved as intended. This did not affect the numerical OpenMolcas/PySCF consistency check or the correctness of the implementation, but it changed the chemical interpretation of the RAS spaces.

I updated `examples/mcscf/gasci_openmolcas/data/al2feo4_plus_rasci_mo.npz` with the corrected reference orbitals so that the RAS partition now matches the intended assignment: O 2p in RAS1, Fe 3d in RAS2, and Fe 4d in RAS3. With the updated file, the OpenMolcas/PySCF agreement is:

```text
E(OpenMolcas) = -2045.013598850000 Eh
E(PySCF)      = -2045.013598849092 Eh
diff          = 9.081e-10 Eh
<S^2>         = 8.75000000
```

I also checked the MO shapes and ordering for the other examples under `examples/mcscf/gasci_openmolcas/` and did not find any similar issue.

yideng2025:
I simplified the binomial lookup by replacing the runtime-initialized table with a static read-only table in `2c540a7`. The previous `stdatomic.h`-based initialization was correct, but unnecessary for a small fixed `C(n,k)` table. The new version keeps the same branchless fixed-stride lookup,

```c
binom_tab[(n << 5) + k]
```

while removing `stdatomic.h`, `binom_state`, and `init_binom()`. I rebuilt `libfci_gas.so` and reran the GASCI tests/examples; `test_gasci.py` passes all 20 tests.

yideng2025:
I made a small style-only cleanup for the GAS C/Python comments and spacing in `9f7520a`. This updates the wording in the contraction-plan comments, fixes a few missing or extra blank lines, and normalizes one over-indented block in `fci_contract_gas.c`. No logic was changed. `test_gasci.py` passes all 20 tests.

MatthewRHermes (approved):
Looks good to me!

PR #206, opened 2026-08-13
Spin-orbit interactions via the State Interaction Framework for MCSCF Methods

JangidBhavnesh:
In this PR, I have added the SISO framework for MCSCF methods such as `CASSCF` and `L-PDFT`, to consider the spin-orbit effects via the two step approach. 

- This includes the spin-orbit integral generation using the `BP` or `DKH1` within the `AMFI` approach (as available in OpenMolcas).
- Constructing the `SO` Hamiltonian in the model-space of CAS or L-PDFT states and diagonalizations to get the SO energies and states
- Added functions for the analysis of decomposition of `SISO` states and also computing the $$J$$-values or $$\Omega$$-values.
- Furthermore, added functionalities to use the required operator expressed in `SF` or `SO` basis and data from `SISO` calculation to generate the `SINGLEANISO` datafile, which can be used to compute the magnetic properties (like D-values, g-Tensors etc) 
- Added the examples for implemented functionalities.
- Added unit-tests to cover most of the implemented functions.

**Reference:** J. Chem. Theory Comput. (2026) 22 (1): 318–333

Let me know for any changes.

Thanks.


examples/siso/00-qdptsoc_for_cas.py
```diff
@@ -0,0 +1,94 @@
...
+CASCI energy for each state
+  State 0 weight 0.333333  E = -242.329497518035 S^2 = 0.7500000
+  State 1 weight 0.333333  E = -242.329497505989 S^2 = 0.7500000
+  State 2 weight 0.333333  E = -242.329497505961 S^2 = 0.7500000
+'''
+
+# 4. State interaction
+# Note, the model-space for the SA-CASSCF and SISO should be the same. You can also define the different model-space
+# for the SA-CASSCF and SISO, but then you need to reconstruct your mc object. There are two Hamiltonian options for
+# the SOC calculations: Breit-Pauli (BP) and Douglas-Kroll-Hess (DKH).
```
MatthewRHermes: If this is true, then why does the `SISO` constructor require you to specify the model space? Why can't it just read it off `mc`?

examples/siso/01-qdptsoc_for_mcpdft.py
```diff
@@ -0,0 +1,156 @@
...
+ State 4 Total Energy = -341.5970405174 S^2 = 0.75
+ State 5 Total Energy = -341.5970399780 S^2 = 0.75
+ State 6 Total Energy = -341.5630640073 S^2 = 0.75
+ State 7 Total Energy = -341.5630639026 S^2 = 0.75
+ State 8 Total Energy = -341.5630638737 S^2 = 0.75
+'''
+
+# Now substite the SA-CASSCF energies with the NEVPT2 energies and do the
+# state interaction calculation.
+mc.e_states[:] = doubletEnergies + quartetEnergies
```
MatthewRHermes: Can this be automated?

examples/siso/00-qdptsoc_for_cas.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Note: to select the model-space which is N_i states of SM_i spin multiplicity,
+# you need to define the tuple in this format [(N_i, SM_i,), (N_j, SM_j), ...].
+# Additionally, you can also specify the symmetry of the target states
+# as [(N_i, SM_i, wfnsym_i), (N_j, SM_j, wfnsym_j), ...].
+# If the symmetry is not specified, it will be set to None by default.
+
+# 3. State-average Calculation.
+mc = mcscf.CASSCF(mf, 4, 3)
+mc = siso.sacasscf_solver(mc, [(3, 2),]) # Here, I am defining a model space of 3 doublets. (N_i=3, SM_i=2): (2P States)
```
MatthewRHermes, line 60: In all these examples, is it _necessary_ to use this function, or is this just syntactic sugar for quickly building the appropriate SA-CASSCF/L-PDFT object? Does this support a `CASCI` object with `nroots>1`?

pyscf/siso/test/test_siso_c_contract.py
```diff
@@ -0,0 +1,133 @@
...
+                            for linkb in link_indexb[strb]:
+                                if spin_change > 0:
+                                    p, q = linkb[0], linka[1]
+                                else:
+                                    p, q = linka[0], linkb[1]
+                                value += (linka[3] * linkb[3]
+                                          * h1e[component, p, q]
+                                          * ci0[root, linka[2], linkb[2]])
+                    ci1[component, root, stra, strb] = value
+    return ci1
```
MatthewRHermes, line 78: If you messed up a sign in the c code, you probably also messed it up here. Have you considered trying to use dummy orbitals to test this?

pyscf/siso/addons.py
```diff
@@ -0,0 +1,492 @@
...
+            zip(energies, lambda_values)):
+        log.note('{:>3d} {:>22.10f} {:>16.4f} {:>8d}'.format(
+            state,
+            energy,
+            lambda_value,
+            int(block_number_by_state[state])))
+
+    return projection_matrix, lambda_values
+
+def soc_state_analysis(siso_data, log, state=(0,), threshold=5e-3):
```
MatthewRHermes, line 391: Docstring should also describe the return values as well as the logger output.

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: I think it's easier to pass on a tuple rather than regenerating it from the mc object. Therefore from API point of view I passed the model-space.

MatthewRHermes: Easier for the developer, or the user? What happens when you pass a tuple that doesn't agree with the state of the `mc` object?

pyscf/siso/anisoaddons.py
```diff
@@ -0,0 +1,555 @@
...
+
+def _basis_transformation(mat, sivec):
+    '''
+    Convert the given matrix to spin orbit basis.
+    '''
+    return np.array([reduce(np.dot, (sivec.conj().T, m, sivec)) for m in mat])
+
+def get_twice_ms_values(mult):
+    """Return integer ``2*M_S`` values for a spin multiplicity."""
+    return list(range(1 - mult, mult, 2))
```
MatthewRHermes: you don't need this. you can just write `list(range(1 - mult, mult, 2))`

pyscf/siso/siso.py
```diff
@@ -0,0 +1,722 @@
...
+
+    g = np.zeros((3, sbra, sket), dtype='complex')
+
+    for g0 in range(3):
+        for g1 in range(sbra):
+            phase = (-1) ** (g0 + S / 2 - (g1 - S / 2))
+            for g2 in range(sket):
+                g[g0, g1, g2] += phase * cg.Wigner3j(ss/2, -(g1 - ss/2),
+                                                     1, 1 - g0,
+                                                     ss/2 + Ms, g2 - ss/2 - Ms).subs(ss,S).doit()
```
MatthewRHermes: why can't you just directly use `cg.CG`? why do you have to compute the phase yourself?

pyscf/siso/socaddons.py
```diff
@@ -0,0 +1,278 @@
...
+
+    if soc1e and soc2e:
+        hso = hso1e+hso2e
+    elif soc1e:
+        hso = hso1e
+    elif soc2e:
+        hso = hso2e
+    return np.array([x.T for x in hso])
+
+def sacasscf_solver(mc, states, weights=None, ms=None):
```
MatthewRHermes: From a user's point of view, calling something called "sacasscf solver" to set up an L-PDFT calculation is awkward. Could you at least overload it with "lpdft solver" or something?

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: I have [sanity check](https://github.com/JangidBhavnesh/pyscf-forge/blob/c895b50ca64040e50f41249a8003dbcd99951e6c/pyscf/siso/siso.py#L577), it will throw an error in the case model space of fed in mc object and requested during the SISO calculation differ.

examples/siso/01-qdptsoc_for_mcpdft.py
JangidBhavnesh: Added a function to automate this.

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: It's a wrapper to create the various `CSFSolvers` of specific spin-states and wfnsymm. I will also dispatch this function as the `state_average_solver` to avoid confusion for L-PDFT method. 

It does work for the CASCI object with `nroots >= 1`.

pyscf/siso/anisoaddons.py
JangidBhavnesh: Removed this fn. Previously I have used this at more than one places, but after cleanup it's only used once, so remove the function and directly given the code.

Thanks.

pyscf/siso/addons.py
JangidBhavnesh: Updated all the docstr in this file to make sure they have the return values/logger output.

pyscf/siso/socaddons.py
JangidBhavnesh: I have updated it to general `state_average-solver`. Kept the `sacasscf_solver` for backward compatibility.

pyscf/siso/siso.py
JangidBhavnesh: That's true, I can directly use it as well.  Changed it to `cg.CG`

pyscf/siso/test/test_siso_c_contract.py
JangidBhavnesh: Thanks for pointing out this.


Now, I have included an additional reference test that introduces a dummy spatial orbital $$d$$, appended after the physical orbitals, and embeds the bra and ket into a common fixed $$(N_\alpha,N_\beta)$$ electron sector. Then the desired spin-flip one-body transition density can be evaluated as a particular element of an ordinary two-electron transition density using PySCF's generic two-electron contraction.

For the $$S_+$$ sector, the dummy-orbital reference computes

$$\widetilde{\Gamma}^{+}_{pq}=\left\langle\widetilde{\Psi}_I\left|
a_{p\alpha}^{\dagger}
a_{d\beta}^{\dagger}
a_{d\alpha}
a_{q\beta}
\right|
\widetilde{\Psi}_J
\right\rangle$$

where the dummy orbital changes from $$d\alpha$$ in the ket to $$d\beta$$ in the bra. Contracting out the dummy orbital reduces this expression to the physical spin-flip TD,

$$\widetilde{\Gamma}^{+}_{pq}=(-1)^{N_\alpha}\left\langle
\Psi_I
\left|
a_{p\alpha}^{\dagger} a_{q\beta}
\right|
\Psi_J
\right\rangle$$

Similarly, for the $$S_-$$ sector, the dummy-orbital reference computes

$$\widetilde{\Gamma}^{-}_{pq}=
\left\langle
\widetilde{\Psi}_I
\left|
a_{d\alpha}^{\dagger}
a_{p\beta}^{\dagger}
a_{d\beta}
a_{q\alpha}
\right|
\widetilde{\Psi}_J
\right\rangle$$

which reduces to

$$\widetilde{\Gamma}^{-}_{pq}=
(-1)^{N_\alpha-1}
\left\langle
\Psi_I
\left|
a_{p\beta}^{\dagger} a_{q\alpha}
\right|
\Psi_J
\right\rangle$$

Thus, the generic two-electron dummy-orbital reference reproduces the direct \(S_+\) and \(S_-\) contractions up to the fermionic parity factors

$$
S_+:\quad (-1)^{N_\alpha},
\qquad
S_-:\quad (-1)^{N_\alpha-1}.
$$

JangidBhavnesh: The initial test cases happened to lie in sectors for which these factors were ``+1``, so the discrepancy was hidden. Anyway, the phase does not affect the SISO energy spectrum because it only rephases adjacent-spin model-space blocks. It does, however, change the phase convention of the SI eigenvectors relative to PySCF, which can matter when those eigenvectors are combined with constructed operators using PySCF functions.

Therefore, I added the corresponding fermionic parity factors to the low-level C contractions for $$S_+$$ and $$S_-$$, and updated the Python reference implementation in the unit test accordingly.

examples/siso/00-qdptsoc_for_cas.py
```diff
@@ -0,0 +1,95 @@
...
```
MatthewRHermes, line 60: In comments in this file, explain that this function is just syntactic sugar and present the 3-4 lines (or whatever) of standard PySCF code that does what this function does.

examples/siso/00-qdptsoc_for_cas.py
MatthewRHermes: Sorry to be obstinate about this but if you have a sanity check then you already have the code to read the model space configuration from the `mc` object, so why is it easier to force the user to type it in twice?

examples/siso/01-qdptsoc_for_mcpdft_or_nevpt2.py
```diff
@@ -0,0 +1,125 @@
...
+ 14                  0.077167944              2.09985          16936.40611
+ 15                  0.077167944              2.09985          16936.40611
+ 16                  0.077258343              2.10231          16956.24632
+ 17                  0.077258343              2.10231          16956.24632
+ 18                  0.077258692              2.10232          16956.32290
+ 19                  0.077258692              2.10232          16956.32290
+'''
+
+# Compute NEVPT2 energies for the model space and update mc.e_states.
+siso.compute_nevpt2_energies(mc, [(8, 2), (1, 4)])
```
MatthewRHermes: 1. Impure functions which edit the states of their arguments (`mc`) should be named with a trailing underscore.
2. It might be more pedagogical to make this a pure function, collect the energies on the left-hand side, and have one more line `mc.e_states = e_nevpt2`. That would be a compromise between the original, complicated version of this example and this extremely concise version, and it would teach the user about the general idea of _replacing the diagonal elements of the Hamiltonian by editing mc.e_states_.
3. Either way, you need to explain just a bit more to the user what this function does. Does it produce logger output? If you leave it as an impure function, you need to say just a little bit more ("update mc.e_states" how? Hold the user's hand).

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: Thanks for pointing this out.

I have added the commentary on this function and have shown the piece of code that this function is automating.
Edited both `00-qdptsoc_for_cas.py` and `01-qdptsoc_for_mcpdft_*.py`.

examples/siso/01-qdptsoc_for_mcpdft_or_nevpt2.py
JangidBhavnesh: I liked the 2nd option. In the recently modified code, I have kept `compute_nevpt2_energy`  a pure function. After computing the energies, I have shown how the diagonal approximation works in the example file.

Thanks.

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: Okay.

I have added a function that will read the model space directly from the `mc` object. Updated the code, example, and unit test to follow this. 

I have kept the model-space as a `kwargs` for backward compatibility.

examples/siso/00-qdptsoc_for_cas.py
```diff
@@ -0,0 +1,107 @@
...
+# calculation, and the second step is to perform the QDPT-SOC calculation using the
+# state-averaged CASSCF wave functions as the model space.
+
+# If you are using this code, please consider citing: J. Chem. Theory Comput. 2026, 22, 1, 318–333
+
+# 1. Define the molecule
+mol = gto.Mole(atom="Al 0 0 0",
+               spin=1,
+               max_memory=10000,
+               basis="ano@5s4p2d1f", #=ANO_RCC_VTZP
```
MatthewRHermes, line 16: Is 3ζ really necessary in an example script? These seem like really hefty examples. At time of writing, `02-qdptsoc_for_lpdft.py` has been running on my laptop for a half hour, and spent most of that time in disk sleep. I've been watching my scratch directory and there's a scratch file that has grown so far from about 2.5 gb to nearly 8 gb and shows no sign of stopping.

JangidBhavnesh: Previously I was generating the `SOC` integrals via doing the `AO2MO` and contracting the `dm` in MO basis. Currently, I have changed it to PySCF's JK builder. This will avoid the $$\mathcal O (N^4)$$ scaling and would avoid the large disk requirement as well.

Added an unit test to verify that this optimized version of implementation is matching exactly with the previous implementation.

JangidBhavnesh: I checked for the `02-qdptsoc_for_lpdft.py` example,  the time takes by SOC integral generation reduced from `>30 minutes` to less than 2 minutes on my computer.

JangidBhavnesh: I have reduced the basis set size from `ANO-RCC-VTZP` to `ANO-MB`. Now it should be quick.

JangidBhavnesh: For example `04-anisdodata.py`, the cost is due to 21 CAS roots, not due to the basis or the SOC integrals. I am keeping this example as it's benchmark molecule for the magnetic properties computation.

JangidBhavnesh: I have reduce the number of roots to make it quicker.

examples/siso/01-qdptsoc_for_mcpdft_or_nevpt2.py
```diff
@@ -0,0 +1,128 @@
+from pyscf import gto, scf, mcpdft
+from pyscf.mcscf import avas
+from pyscf import siso
+
+# Two-step state interaction calculation with dynamical correlation effects included via MC-PDFT.
+# Alternatively, you can also include the dynamical correlation effects via NEVPT2.
+
+mol = gto.Mole(atom="P 0 0 0",
+               spin=3,
+               max_memory=120000,
```
MatthewRHermes: sorry i didn't notice this earlier, but this calculation does not require anywhere near that much memory and if it did, it wouldn't be a good example.

examples/siso/04-anisodatagen.py
```diff
@@ -0,0 +1,58 @@
...
+# 1. Define the molecule
+mol = gto.Mole()
+mol.atom='''
+Dy 0 0 0
+O 0 0 1.74'''
+mol.basis= {'Dy': 'ano@6s5p3d1f', 'O': 'ano@2s1p'} # ANO-MB
+mol.spin = 5
+mol.charge = 1
+mol.verbose = 4
+mol.max_memory = 120000
```
MatthewRHermes: no. delete `max_memory` line.

examples/siso/01-qdptsoc_for_mcpdft_or_nevpt2.py
JangidBhavnesh: I doesn't use that much mem, reduced the mem to 10 GB.

examples/siso/04-anisodatagen.py
JangidBhavnesh: removed.

examples/siso/01-qdptsoc_for_mcpdft_or_nevpt2.py
JangidBhavnesh: removed the mem line altogether.

examples/siso/00-qdptsoc_for_cas.py
JangidBhavnesh: Changed the example itself.

MatthewRHermes (approved):
tyvm!

PR #210, opened 2026-08-31
Add SF-TDDFT nonadiabatic couplings and batch caching

jingjjkk:
This PR adds analytic nonadiabatic couplings (NACs) for spin-flip TDA/TDDFT and introduces reusable caching and batch utilities for gradient/NAC calculations.

Main changes
- Add spin-flip TDA/TDDFT nonadiabatic coupling implementation in `pyscf.nac.tduks_sf`.
- Support state-pair NAC calculations with optional energy-difference scaling and electron-translation-factor corrections.
- Reuse XC grid data, response intermediates, and J/K evaluations between gradient and NAC calculations.
- Add regression tests comparing the new implementation with existing gradient and NAC reference values.

### DIFF
--- a/pyscf/nac/__init__.py
+++ b/pyscf/nac/__init__.py
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+
+from pyscf.nac.tduks_sf import NAC
+from pyscf.nac.tduks_sf import NonAdiabaticCouplings
+
+__all__ = ['NAC', 'NonAdiabaticCouplings']
--- a/pyscf/nac/__init__.py
+++ b/pyscf/nac/__init__.py
@@ -1,6 +1,32 @@
 #!/usr/bin/env python
+# Copyright 2014-2024 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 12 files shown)


MatthewRHermes:
The `tools` folder is for very generic utilities, I/O, or API. Since you have a bunch of files with `nac` in the name, just put them in the `nac` folder instead.

pyscf/nac/gradient_nac_cache.py
MatthewRHermes, line 1: Is this absolutely necessary? If it is, can you use a context manager instead of a single global variable? And please add unit tests for this functionality specifically.

pyscf/nac/__init__.py
MatthewRHermes, line 1: Please consult and compare PySCF core `pyscf/nac/__init__.py`, which this will have to retain compatibility with. I don't think you should give the generic names `NAC` and `NonAdiabaticCouplings` to the quantities computed at this specific level of theory in this layer.

pyscf/nac/test/test_gradient_nac_cache.py
MatthewRHermes, line 1: Is there a test here that validates that the jk result from the cache is the same as the jk computed naively, without the cache? Is there a test that the end-to-end result of a calculation is unchanged whether the cache is enabled or not?

MatthewRHermes (approved):

PR #211, opened 2026-09-14
Add generalized active space self-consistent field solver

yideng2025:
## Summary

This PR adds a generalized active space self-consistent field (GASSCF) implementation to `pyscf-forge`. The implementation builds on the determinant-based GASCI solver and PySCF's existing second-order CASSCF framework `newton_casscf`. It includes:

* A `GASSCF` driver following the existing PySCF CASSCF interface.
* GAS-aware orbital optimization, including nonredundant inter-GAS active-active rotations.
* Joint orbital/CI second-order optimization through the native PySCF Newton/CIAH machinery.
* State-specific and state-averaged GASSCF, including zero-weight state-average roots.
* GAS-safe core and virtual orbital canonicalization.
* Energy scanner support through `as_scanner()`.
* Spin-penalty support inherited from the GASCI solver.
* Density-fitting support through both `density_fit()` and `DFGASSCF`.
* Unit tests covering GAS-as-CAS limits, genuine restricted-GAS calculations, state averaging, canonicalization, scanners, and density fitting.

### GASSCF driver

The public interface is provided by:

```python
from pyscf.mcscf import gasscf
```

and follows the existing PySCF MCSCF object style:

```python
mc = gasscf.GASSCF(
    mf,
    gas_orbs=(2, 4, 2),
    gas_restr=((2, 4), (7, 9), (10, 10)),
    gas_restr_type="cumulative-occ",
    nelecas=(5, 5),
    ncore=2,
)
mc.kernel()
```

The GAS definition and restriction parsing are shared with `pyscf.mcscf.gasci`. All GAS restriction formats accepted by GASCI are therefore supported by GASSCF, including explicit spin-supergroups, supergroups, cumulative-occupation restrictions, and RAS-style restrictions.

### Orbital optimization

The implementation reuses PySCF's native second-order CASSCF/Newton optimizer rather than introducing a separate orbital-optimization framework. The native augmented-Hessian machinery provides the macro/micro iteration structure, keyframes, trust-region handling, and orbital/CI Newton updates. The GASSCF layer adapts this machinery to the restricted GAS determinant space by providing:

* GASCI vectors and Hamiltonian contractions.
* GAS one- and two-electron RDMs and transition RDMs.
* GAS-aware CI response operations.
* GAS helper-plan lifetime and cache management.
* A GAS orbital-rotation mask.

In addition to the usual core-active, core-virtual, and active-virtual rotations, rotations between different GAS subspaces are included because they are nonredundant in GASSCF. Active-active rotations within the same GAS subspace remain redundant and are excluded. For a single GAS containing the complete active space, the implementation reduces to the ordinary CASSCF orbital-optimization problem and is tested against `newton_casscf`.

### State averaging

Ordinary state-averaged GASSCF is supported through the existing PySCF state-average interface. Zero-weight roots are also supported. Such roots are solved and retained in the CI state space and can affect the optimization path, but do not contribute directly to the weighted state-average objective. The state-specific limit is validated using state-average weights of the form

```python
(1.0, 0.0, ...)
```

and reproduces the corresponding state-specific stationary solution.

### Canonicalization and scanners

GASSCF supports canonicalization of the core and virtual orbital spaces. Active space natural orbital rotations are not applied because rotations mixing different GAS subspaces can change the definition of a restricted GAS wave function.

Energy-only potential energy surface scans are supported through:

```python
scanner = mc.as_scanner()
```

### Density fitting

Density fitting is supported through the usual PySCF-style interface:

```python
mc = gasscf.GASSCF(...)
mc = mc.density_fit()
```

and through the `DFGASSCF` factory. Both construction paths are covered by the numerical regression tests.

## Tests

The implementation is tested at several levels.

### GAS-as-CAS limit

A single GAS containing the complete active determinant space is compared with the existing PySCF `newton_casscf` implementation. This validates the GASSCF adapter and the native Newton driver in the limit where GASSCF reduces to CASSCF.

### Genuine restricted-GAS regression

A converged N2 genuine-GAS calculation is used as a numerical regression test:

```text
ncore     = 2
nelecas   = (5, 5)
gas_orbs  = (2, 4, 2)
gas_restr = ((2, 4), (7, 9), (10, 10))
```

Using converged OpenMolcas GASSCF orbitals, the fixed-orbital GASCI energy is

```text
PySCF      -109.017476499975 Eh
OpenMolcas -109.017476500000 Eh
```

and the state-specific GASSCF result is

```text
PySCF      -109.017476499980 Eh
OpenMolcas -109.017476500000 Eh
```

The difference is approximately `2e-11 Eh`. The same regression fixture is used to validate:

* State-specific GASSCF.
* Zero-weight state averaging.
* Ordinary state averaging.
* Density-fitting entry points.
* Energy scanning.
* Canonicalization invariance.

The zero-weight state-average result agrees with the state-specific energy to approximately `3e-11 Eh`. Core/virtual canonicalization preserves the fixed-orbital GAS energy to approximately `2e-12 Eh`.

### Unit tests

The current test suites pass with:

```text
pyscf.mcscf.test.test_gasscf: 59 tests passed
pyscf.mcscf.test.test_gasci:  23 tests passed
```

They can be run with:

```bash
python -m unittest pyscf.mcscf.test.test_gasscf
python -m unittest pyscf.mcscf.test.test_gasci
```

## Scope

This PR is intentionally limited to GASSCF orbital optimization based on the existing determinant-based GASCI solver.

It does not add:

* State-average-mix GASSCF.
* State-specific excited-state wrappers.
* Analytic nuclear gradients or nonadiabatic couplings.
* The legacy two-step CASSCF driver.

### DIFF
--- a/pyscf/mcscf/gasscf.py
+++ b/pyscf/mcscf/gasscf.py
@@ -27,8 +27,9 @@ Supported GAS definitions follow :mod:`pyscf.mcscf.gasci`: ``gas_orbs``,
 ``gas_restr`` and ``gas_restr_type`` are normalized by the same GAS helper
 routines.  This module supports ordinary state averaging, including
 zero-weight roots, GAS-safe canonicalization of inactive/external orbitals,
-energy-only scanners and the GASCI-native spin-penalty Hamiltonian.  It does
-not implement state-average-mix, state-specific excited-state wrappers,
+density fitting through PySCF's ``mcscf.df`` machinery, energy-only scanners
+and the GASCI-native spin-penalty Hamiltonian.  It does not implement
+state-average-mix, state-specific excited-state wrappers,
 active-space natural-orbital rotations, analytic gradients/NACs or the
 legacy two-step CASSCF driver.
 """
@@ -46,11 +47,12 @@ from pyscf.lib import logger
 from pyscf.fci import addons as fci_addons
 from pyscf.mcscf import addons
 from pyscf.mcscf import addons_gas
+from pyscf.mcscf import df as mcdf
 from pyscf.mcscf import fci_gas
 from pyscf.mcscf import gasci
 from pyscf.mcscf import newton_casscf
 
-__all__ = ["GASSCF"]
+__all__ = ["GASSCF", "DFGASSCF"]
 
 
 def _unsupported(feature):
@@ -659,6 +661,29 @@ class _StateAverageGASSCF(addons.StateAverageMCSCF):
         result.fcisolver.nroots = 1
         result.fcisolver._init_plan_cache()
         result.fcisolver.mol = result.mol
+        return result
+
+
+class _DFGASSCF(mcdf._DFCASSCF):
+    """Density-fitting mixin with GAS-specific user-visible labels."""
+
+    __name_mixin__ = "DF"
+
+    def dump_flags(self, verbose=None):
+        super(mcdf._DFCAS, self).dump_flags(verbose)
+        logger.info(
+            self,
+            "DFGASSCF: density fitting for JK matrix and 2e integral "
+            "transformation")
+        return self
+
+    def undo_df(self):
+        self.close()
+        result = lib.view(self, lib.drop_class(self.__class__, _DFGASSCF))
+        try:
+            del result.with_df
+        except AttributeError:
+            pass
         return result
 
 
@@ -1329,6 +1354,22 @@ class GASSCF(newton_casscf.CASSCF):
             "method": getattr(self.fcisolver, "spin_penalty_method", None),
         }
 
+    def density_fit(self, auxbasis=None, with_df=None):
+        """Return a DF-GASSCF object using PySCF's CASSCF DF machinery.
+
+        Density fitting changes only integral/J-K construction.  The active
+        CI vector, GAS restrictions, GAS RDMs and spin-penalty bookkeeping
+        remain owned by the GASSCF/GASCI adapter.
+        """
+
+        result = mcdf.density_fit(self, auxbasis=auxbasis, with_df=with_df)
+        if isinstance(result, _DFGASSCF):
+            return result
+        if isinstance(result, mcdf._DFCASSCF):
+            result.__class__ = lib.replace_class(
+                result.__class__, mcdf._DFCASSCF, _DFGASSCF)
+        return result
+
 
     def as_scanner(self):
         """Return an energy-only scanner for a fixed GASSCF objective."""
@@ -1431,3 +1472,19 @@ class GASSCF(newton_casscf.CASSCF):
         """Alias for the inherited active-space energy slot."""
 
         return self.e_cas
+
+
+def DFGASSCF(mf, gas_orbs=None, gas_restr=None, *, nelecas,
+             gas_restr_type=None, ncore=None, frozen=None, fcisolver=None,
+             cache_plans=None, auxbasis=None, with_df=None):
+    """Create a density-fitted :class:`GASSCF` object.
+
+    This mirrors PySCF's ``DFCASCI/DFCASSCF`` construction style while keeping
+    the user-facing GAS interface identical to :class:`GASSCF`.
+    """
+
+    return GASSCF(
+        mf, gas_orbs=gas_orbs, gas_restr=gas_restr,
+        nelecas=nelecas, gas_restr_type=gas_restr_type, ncore=ncore,
+        frozen=frozen, fcisolver=fcisolver,
+        cache_plans=cache_plans).density_fit(auxbasis=auxbasis, with_df=with_df)
--- a/pyscf/mcscf/gasscf.py
+++ b/pyscf/mcscf/gasscf.py
@@ -20,22 +20,24 @@
 
 This module is introduced in small reviewable stages.  Current commits define
 object construction, explicit GASCI solver adaptation, GAS orbital-rotation
-masks, Newton-owned GAS helper plan lifetimes, public solver dispatch through
+masks, GASSCF-owned GAS helper plan lifetimes, public solver dispatch through
 those plans, GASCI-like object-level wrappers, a fixed-orbital GASCI bridge,
 and a minimal native Newton/CIAH driver bridge.  Current validation also
 guards unsupported staged feature combinations, kernel lifetimes and
-ordinary state-average wrappers, GAS-safe canonicalization, energy scanner support
-and GASCI-native spin-penalty hooks.
+ordinary state-average wrappers, GAS-safe canonicalization, energy scanner support,
+GASCI-native spin-penalty hooks and GAS-labeled native driver output.
 """
 
 from collections import OrderedDict
 import hashlib
 import json
+import sys
 
 import numpy
 
 from pyscf import gto
 from pyscf import lib
+from pyscf.lib import logger
 from pyscf.fci import addons as fci_addons
 from pyscf.mcscf import addons
 from pyscf.mcscf import addons_gas
@@ -47,14 +49,85 @@ __all__ = ["GASSCF"]
 
 
 def _unsupported(feature):
-    raise NotImplementedError(feature + " is not implemented for Newton GASSCF")
+    raise NotImplementedError(feature + " is not implemented for GASSCF")
+
+
+
+class _GASSCFLogFilter:
+    """Write-through stream filter for native Newton/CASSCF messages.
+
+    GASSCF deliberately reuses PySCF's native ``newton_casscf`` driver.
+    The numerical driver still contains CASSCF/CASCI text labels and one
+    CASSCF-specific experimental-feature warning.  This filter changes only
+    user-visible text while leaving the driver and all numerical data untouched.
+    """
+
+    _DROP_SUBSTRINGS = (
+        "SO-CASSCF (Second order CASSCF) is an experimental feature. "
+        "Its performance is bad for large systems.",
+    )
+    _REPLACEMENTS = (
+        ("Start SO-CASSCF (newton CASSCF)", "Start SO-GASSCF"),
+        ("newton CASSCF", "GASSCF"),
+        ("Second order CASSCF", "Second order GASSCF"),
+        ("SO-CASSCF", "SO-GASSCF"),
+        ("CASSCF", "GASSCF"),
+        ("CASCI", "GASCI"),
+        ("CAS (", "GAS ("),
+        ("CAS space", "GAS active space"),
+        ("CAS-space", "GAS active-space"),
+        ("E(CI)", "E(GASCI)"),
+    )
+
+    def __init__(self, stream):
+        self._stream = sys.stdout if stream is None else stream
+        self._pending = ""
+
+    def _rewrite_line(self, line):
+        if any(text in line for text in self._DROP_SUBSTRINGS):
+            return ""
+        if "SO-CASSCF" in line and "experimental feature" in line:
+            return ""
+        for old, new in self._REPLACEMENTS:
+            line = line.replace(old, new)
+        return line
+
+    def write(self, text):
+        original_length = len(text)
+        text = self._pending + text
+        if not text:
+            return original_length
+        if text.endswith("\n"):
+            self._pending = ""
+            lines = text.splitlines(True)
+        else:
+            lines = text.splitlines(True)
+            if lines and not lines[-1].endswith("\n"):
+                self._pending = lines.pop()
+            else:
+                self._pending = ""
+        rewritten = "".join(self._rewrite_line(line) for line in lines)
+        if rewritten:
+            self._stream.write(rewritten)
+        return original_length
+
+    def flush(self):
+        if self._pending:
+            rewritten = self._rewrite_line(self._pending)
+            self._pending = ""
+            if rewritten:
+                self._stream.write(rewritten)
+        return self._stream.flush()
+
+    def __getattr__(self, name):
+        return getattr(self._stream, name)
 
 
 class _GASFCISolver(fci_gas.FCISolver):
-    """GASCI solver shell with Newton GASSCF-owned helper plans.
+    """GASCI solver shell with GASSCF-owned helper plans.
 
     Ordinary GASCI remains implemented by :mod:`fci_gas`.  This subclass owns
-    only reusable helper plans needed by the staged Newton orbital optimizer.
+    only reusable helper plans needed by the staged GASSCF orbital optimizer.
     Adapted or copied solvers always start with empty caches, so C workspaces
     are never borrowed across solver objects.
     """
@@ -585,7 +658,7 @@ class _StateAverageGASSCF(addons.StateAverageMCSCF):
 
 
 class GASSCF(newton_casscf.CASSCF):
-    """Joint Newton orbital optimizer for a determinant GASCI active space.
+    """Joint GASSCF orbital optimizer for a determinant GASCI active space.
 
     Args:
         mf : SCF object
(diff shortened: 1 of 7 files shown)


pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1542 @@
...
+            raise ValueError(
+                "GAS definition, roots, weights or frozen orbitals changed; "
+                "create a new scanner")
+
+        old_mol = self.mol
+        previous_mo = None if self.mo_coeff is None else numpy.array(
+            self.mo_coeff, copy=True)
+        guess_ci = _copy_ci(self.ci if ci0 is None else ci0, signature)
+
+        self.reset(mol)
```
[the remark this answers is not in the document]
yideng2025: `_DFGASSCF` inherits PySCF's `_DFCASSCF` / `_DFCAS`, and `_DFCAS.reset()` already calls `self.with_df.reset(mol)` before forwarding to `super().reset(mol)`. Therefore the auxiliary DF object is already rebound to the new molecule when the scanner calls `self.reset(mol)`. I left the reset implementation unchanged, but added `test_df_scanner_reset_updates_with_df_molecule` in `3c9a0ef` to make this behavior explicit and guard it against regression.

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1542 @@
...
+        result = super().copy()
+        result._init_plan_cache()
+        return result
+
+
+def _new_gas_solver(mf, gas_orbs, gas_restr, gas_restr_type, cache_plans):
+    if gas_orbs is None:
+        raise ValueError(
+            "gas_orbs is required without an explicit GASCI solver")
+    gas_orbs = tuple(int(value) for value in gas_orbs)
```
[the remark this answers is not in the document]
yideng2025: Addressed in `3c9a0ef`. GASSCF now uses a shared `_validated_user_gas_orbs` helper that delegates to the same strict integer validation used by GASCI (`addons_gas._integer_vector`), instead of eagerly converting entries with `int`. The constructor, explicit-solver adaptation path, and public `gas_orbs` setter now all reject floating-point and Boolean values. A regression test covers both `(2.5,)` and `(True,)`.

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1542 @@
...
+    if any(value <= 0 for value in gas_orbs):
+        raise ValueError("gas_orbs entries must be positive integers")
```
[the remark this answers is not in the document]
yideng2025: Addressed in `3c9a0ef`. I removed the GASSCF-specific requirement that every `gas_orbs` entry be strictly positive. GASSCF now preserves the user GAS definition and delegates the restriction semantics to the shared GASCI normalizer. This restores support for valid RAS specifications with empty RAS1/RAS3 spaces, such as `gas_orbs=(0, 2, 0)`. The new regression test checks both the normal constructor and explicit-GASCI-solver construction paths and verifies that the kernel space is normalized to `(2,)`.

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1542 @@
...
+
+        stdout = getattr(self, "stdout", None)
+        restore_stdout = not isinstance(stdout, _GASSCFLogFilter)
+        if restore_stdout:
+            self.stdout = _GASSCFLogFilter(stdout)
+
+        stderr = sys.stderr
+        restore_stderr = not isinstance(stderr, _GASSCFLogFilter)
+        if restore_stderr:
+            sys.stderr = _GASSCFLogFilter(stderr)
```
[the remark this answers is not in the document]
yideng2025: Addressed in `3c9a0ef`. GASSCF no longer replaces process-global `sys.stderr`. The native Newton logging is now handled with a calculation-local `_GASSCFLogger` together with the existing stream-label filter. The logger suppresses the inherited SO-CASSCF experimental warning without mutating global process state. A regression test also checks that `sys.stderr` remains unchanged while the warning is filtered.

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1542 @@
...
+            Number of inactive doubly occupied orbitals.
+        frozen : int or sequence of ints, optional
+            Frozen orbital specification forwarded to the native Newton class.
+        cache_plans : bool, optional
+            Future policy for Newton-owned GAS contraction/RDM/spin plans.
+
+    Notes:
+        This first implementation stage establishes the public object and
+        solver ownership convention only.  Later commits add the GAS orbital
+        rotation mask and native CIAH derivative adapter.
```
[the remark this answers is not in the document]
yideng2025: Addressed in `3c9a0ef`. I updated the public `GASSCF` docstring and the remaining development-stage comments so they describe the current implementation: the native PySCF Newton/CIAH driver is reused, with GAS-specific CI/RDM/spin operations and the GAS orbital-rotation mask supplied by the adapter. The outdated references to later/staged implementation work have been removed.

yideng2025:
Addressed the review findings in `3c9a0ef`, including strict GAS input validation, zero-sized RAS spaces, calculation-local logging, native Newton ERIS reuse, scanner orbital-mask invariants, and stale documentation. Added targeted regression coverage; the GASSCF and GASCI test suites pass 64/64 and 23/23 tests respectively.

yideng2025:
A short note on why this implementation is based on `newton_casscf` rather than the more commonly used `mc1step` CASSCF driver.

During development, we first explored a GASSCF extension of the `mc1step`/CIAH framework. This works well for ordinary CASSCF because much of the orbital optimization can be described by the fixed-CI orbital Hessian, with the CI wave function updated co-iteratively. For GASSCF, however, rotations between different GAS subspaces are nonredundant and can couple strongly to changes in the restricted CI wave function. In the systems we tested, this CI response was important for the orbital curvature and could not be neglected without degrading convergence.

We therefore tested several `mc1step`-based variants, including the fixed-CI Hessian and explicit CI-response corrections. The fixed-CI approximation could have difficulty converging for strongly coupled cases, while treating the CI response explicitly introduced a substantially more expensive nested response problem and was difficult to integrate robustly into the CIAH microiterations.

This motivated the final design used in this PR: reuse PySCF's existing `newton_casscf` framework and solve the orbital and CI variables together in the joint augmented-Hessian/Newton problem. In the GAS-as-CAS limit this reduces to the existing PySCF second-order CASSCF formulation, while for genuine GAS spaces the additional inter-GAS orbital variables and GASCI response are handled within the same joint Newton space.

The intent here is therefore not to replace `mc1step` as the standard CASSCF optimizer, but to use the existing joint orbital/CI Newton machinery because it provides a more natural and robust treatment of the stronger CI response encountered in GASSCF.

pyscf/mcscf/test/test_gasscf.py
```diff
@@ -0,0 +1,2417 @@
...
+            delta=self.N2_REGRESSION_TOL)
+
+
+    def test_public_gasscf_module_exports_gasscf(self):
+        from pyscf.mcscf import gasscf as imported_gasscf
+
+        self.assertIs(imported_gasscf.GASSCF, gasscf.GASSCF)
+        self.assertEqual(imported_gasscf.__all__, ["GASSCF", "DFGASSCF"])
+        self.assertIs(imported_gasscf.DFGASSCF, gasscf.DFGASSCF)
+        self.assertEqual(gasscf.GASSCF.__module__, "pyscf.mcscf.gasscf")
```
MatthewRHermes: what could ever possibly cause this test to fail?

pyscf/mcscf/test/test_gasscf.py
```diff
@@ -0,0 +1,2417 @@
...
+    def test_natural_orbital_rotations_are_guarded(self):
+        mol = gto.M(atom="H 0 0 0; H 0 0 0.75", basis="sto-3g", verbose=0)
+        mf = scf.RHF(mol).run()
+        mc = gasscf.GASSCF(
+            mf, gas_orbs=(2,), gas_restr=None, nelecas=(1, 1), ncore=0)
+
+        with self.assertRaisesRegex(NotImplementedError, "natural-orbital"):
+            mc.canonicalize(mf.mo_coeff, gas_natorb=True)
+        with self.assertRaisesRegex(NotImplementedError, "natural-orbital"):
+            mc.cas_natorb()
```
MatthewRHermes, line 2118: I understand that we can't support, i.e., `cas_natorb_` here that updates the `mc.mo_coeff` attribute, because it will mix orbitals from different subspaces and it breaks the CI vector. But there still should be some support for presenting the natural orbitals for external visualization via molden.

pyscf/mcscf/test/test_gasscf.py
yideng2025: Good point. This test was a leftover from an earlier stage of the implementation.

Originally, the orbital optimizer lived in `newton_gasscf.py`, following the role of PySCF's `newton_casscf.py`. Since pyscf-forge should not modify the core PySCF package `__init__.py`, we introduced `gasscf.py` as the public-facing module and used this test to check that the public `GASSCF`/`DFGASSCF` symbols were wired to the intended implementation classes.

As the implementation matured, we decided not to maintain parallel GASSCF orbital-optimization drivers analogous to the different CASSCF drivers such as `mc1step`. The Newton-based implementation became the sole GASSCF optimizer, so the implementation was moved into `gasscf.py` itself.

At that point this test became redundant: the public module and the implementation module are now the same module, and `test_gasscf.py` already imports `pyscf.mcscf.gasscf` at module scope. The identity checks therefore no longer provide an independent regression test.

I removed the test in `400b812`. The actual `GASSCF` and `DFGASSCF` constructors and their behavior continue to be exercised throughout the rest of the test suite.

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1559 @@
...
+        explicitly guarded.  The active GAS block is otherwise kept fixed;
+        only inactive and external orbitals are canonicalized by the inherited
+        PySCF machinery.
+        """
+
+        if gas_natorb or cas_natorb:
+            _unsupported("GAS natural-orbital rotation")
+        return gasci.GASCI.canonicalize(
+            self, mo_coeff, ci, eris, sort=sort, gas_natorb=False,
+            gasdm1=gasdm1, verbose=verbose, **kwargs)
```
MatthewRHermes, line 1250: What do you think about the possibility of implementing the CI vector rotations on `get_gas_pseudo_natorb`, and then including that as an option here?

pyscf/mcscf/gasscf.py
```diff
@@ -0,0 +1,1559 @@
...
+        """Return an energy-only scanner for a fixed GASSCF objective."""
+
+        return _as_scanner(self)
+
+    def state_specific_(self, *args, **kwargs):
+        _unsupported("state-specific GASSCF")
+
+    state_specific = state_specific_
+
+    gen_g_hop = newton_casscf.gen_g_hop
```
MatthewRHermes, line 1458: We need a unit test for this specifically, showing that the relative error against finite difference with step size $\delta$ goes to zero as $O(\delta^1)$ in $\lim_{\delta\to 0}$. At least for the gradient, and with non-trivial GAS subspaces.

pyscf/mcscf/test/test_gasscf.py
yideng2025: Addressed in `f2b92b7`. GASSCF now reuses GASCI's `get_gas_natorb`, `get_gas_average_natorb`, and `get_gas_pseudo_natorb` interfaces. These return orbital coefficients and occupations for analysis without modifying `mc.mo_coeff` or the CI vectors.

I added a `molden.from_mo()` example to the GASSCF docstring, including how to assemble the full occupation array from the core, active, and virtual contributions. A regression test writes and reads back Molden files for root-specific natural orbitals, state-averaged natural orbitals, and GAS pseudo-natural orbitals, checking the coefficients and occupations.

pyscf/mcscf/gasscf.py
yideng2025: Implemented in `1ec5d4f` via `canonicalize(gas_pseudo_natorb=True)`. It diagonalizes each GAS density block and transforms every CI root, including zero-weight roots. SA uses the weighted density by default. `canonicalize` returns the results; `canonicalize_` writes them back. The option defaults to `False`, and `get_gas_pseudo_natorb` remains analysis-only.

Mathematically, orbital rotations induce determinant transformations through minors of the orbital rotation matrix. We evaluate the same transformation by factoring each GAS orbital block into Givens rotations between adjacent orbitals, plus diagonal sign factors. Each rotation mixes only the corresponding singly occupied string pairs, preserving GAS occupations and avoiding full-CAS expansion. This does not change Newton's orbital-update algorithm.

Tests against full-FCI transformations and N2 calculations verify preservation of root energies, AO densities, and overlaps, including DF and zero-weight SA.

pyscf/mcscf/gasscf.py
yideng2025: Added in `201d0cb` using the embedded N2 fixture with a restricted GAS(2,4,2) space. A small deterministic orbital rotation moves the fixture away from the near-stationary solution.

The tests compare analytical orbital directional derivatives against centered energy differences at fixed CI, using step sizes 2e-4 and 1e-4. They check absolute agreement and error reduction consistent with O(delta^2) truncation, allowing for numerical noise.

Coverage includes core–active, core–virtual, active–virtual, inter-GAS, and combined directions for single-state, SA, zero-weight SA, and DF calculations.

pyscf/mcscf/test/test_gasscf.py
```diff
@@ -0,0 +1,3118 @@
...
+        report = mc.spin_energy_report()
+
+        self.assertTrue(numpy.isfinite(e_tot))
+        self.assertTrue(numpy.isfinite(e_gas))
+        self.assertEqual(numpy.asarray(ci).ndim, 1)
+        self.assertEqual(mo_coeff.shape, mf.mo_coeff.shape)
+        self.assertIsNone(mo_energy)
+        self.assertTrue(numpy.isfinite(report["physical"]))
+        self.assertTrue(numpy.isfinite(report["penalty"]))
+        self.assertAlmostEqual(report["objective"], e_tot, places=9)
```
MatthewRHermes: I'm think the expectation of users is going to be that `e_tot` follows `report["physical"]`, not `report["objective"]`. Am I reading this correctly?

MatthewRHermes: Similarly, `mc.e_states` should correspond to `report["physical"]` in the context of a state-averaged calculation (although that would rarely even make sense to do, since as we've discussed, state-averaged methods and spin penalties don't really work together all that well).

examples/mcscf/gasscf_openmolcas/04_alfe2o4_plus_fix_spin.py
```diff
@@ -0,0 +1,121 @@
...
+target_s2 = 20.0  # S=4, multiplicity 9.
+
+here = Path(__file__).resolve().parent
+data_file = here / "gasscf_openmolcas/data/alfe2o4_plus_mo.npz"
+
+with numpy.load(data_file, allow_pickle=False) as data:
+    mol = gto.loads(data["mol"].item())
+    mo_coeff = data["mo_coeff"]
+
+mol.max_memory = 16000  # MB
```
MatthewRHermes, line 65: this is how much total memory my laptop has. this whole file is kind of intense for an example script.

pyscf/mcscf/test/test_gasscf.py
yideng2025: Addressed in `74441a0`. Both GASCI and GASSCF now expose physical energies through `e_tot` and the public `kernel()` return values. For state averaging, `e_states` contains each root’s physical energy, and `e_tot` is their weighted average.

The underlying solver and Newton driver retain the penalized objective internally; `spin_energy_report()` provides the physical energy, penalty, and objective separately. Tests check nonzero penalties against independent density matrix energy evaluations, including state-averaged and zero-weight cases.

examples/mcscf/gasscf_openmolcas/04_alfe2o4_plus_fix_spin.py
yideng2025: Addressed in `d24ae9b`. The introductory example is now `93_gasscf_h4.py`, a self-contained H4/STO-3G calculation using a restricted GAS space and default optimization settings, without external data or a spin penalty. The larger AlFe2O4+ calculation has been moved to `gasscf_openmolcas/04_alfe2o4_plus_fix_spin.py` alongside the other OpenMolcas comparisons.

yideng2025:
I have completed a final pass over the current GASSCF implementation and the associated GASCI interfaces. The implementation now has the following structure:

- GASSCF reuses PySCF's native `newton_casscf` joint orbital/CI Newton machinery, with GAS-specific CI contractions, RDM/transition-RDM operations, spin operations, helper-plan management, and the nonredundant inter-GAS orbital rotations supplied by the GAS adapter.
    
- The GAS definition and normalization are shared with GASCI, including `spin-supergroup`, `supergroup`, `cumulative-occ`, and `ras` restrictions.
    
- Ordinary state-specific and state-averaged GASSCF are supported, including zero-weight roots.
    
- Core/virtual canonicalization is supported, together with optional within-GAS pseudo-natural orbital rotations that transform the CI vectors consistently.
    
- Density fitting, energy scanners, and the GASCI-native spin-penalty Hamiltonian are supported.
    
- Public `e_tot`, `e_gas`, and state energies report the physical energies. When a spin penalty is used, the penalized quantity remains internal to the eigensolver/Newton optimization and is exposed separately through `spin_energy_report()`.
    

The review feedback has also led to several robustness improvements. In particular, the current code has strict GAS input validation, support for valid zero-sized RAS edge spaces, calculation-local logging without global stream mutation, explicit ownership and lifetime rules for reusable GAS contraction/RDM/spin plans, scanner and DF resource handling, and GAS-preserving CI transformations for pseudo-natural orbital rotations.

The regression coverage now includes the GAS-as-CAS limit, genuine restricted-GAS calculations, state averaging and zero-weight roots, DF and scanner paths, canonicalization invariance, spin-penalty physical/objective energy separation, helper-plan ownership/lifetime, GAS CI orbital transformations against full-FCI references, and finite-difference checks of the orbital gradient including inter-GAS rotations. The OpenMolcas comparison examples are kept separately, while `93_gasscf_h4.py` provides a small self-contained introductory GASSCF example.

Unsupported features are explicitly guarded rather than falling through to incompatible CASSCF implementations, including state-average-mix/state-specific excited-state wrappers, spatial symmetry, X2C, solvent models, analytic gradients/NACs, GPU conversion, and the legacy CASSCF optimization drivers.

The commit `2904193` is a documentation/comment cleanup and does not change the numerical implementation.

MatthewRHermes (approved):

